PR #1, opened 2020-06-16
Init

ajlee21:
This PR initializes an analysis where we are using our simulation approach (ponyo) to identify genes and pathways that are "generic" (i.e. those that have a higher likelihood to be deferentially expressed across a range of different experiments).

This PR is currently pip installing the ponyo module manually until ponyo is uploaded into pypi

The main parts of this PR is the notebook found in ```human_analysis``` which contains a notebook that processes the recount data and trains a VAE.

PR #2, opened 2020-06-18
Import modules locally

ajlee21:
This PR adds recommended changes from https://github.com/greenelab/generic-expression-patterns/pull/1#discussion_r441555209

This PR adds a setup.py file and imports modules locally using this.


ajlee21:
Good point! I have updated the name to match the repo. so its a little long, but at least is clear.

PR #5, opened 2020-06-25
Test linter on nbconverted files

ajlee21:


ajlee21:
Linter results seem to produce a lot of errors due to ipython magic. Perhaps this can be an alternative for notebook formatting: https://neuralcoder.science/Black-Jupyter/

PR #7, opened 2020-07-10
Add pseudomonas analysis

ajlee21:
This PR runs a new dataset, a pseudomonas dataset, through the same analysis pipelines to identify potentially generic genes.

The main changes include:
1. Adding a new notebook to process the pseudomonas data, `pseudomonas_analysis/1_process_pseudomonas_data.ipynb`. This notebook is different then the processing notebook for recount2 due to differences in the dataset.

2. Adding notebook to identify generic genes, `pseudomonas_analysis/2identify_generic_genes_pathways.ipynb`. This notebook is essentially redundant to `human_analysis/2identify_generic_genes_pathways.ipynb`. I'm not sure if this is the best way to handle this pipeline but I didn't want to overwrite the results from the human analysis.

3. Updated the readme and modules to make running this analysis on a different dataset easier. Additional suggestions on this front are very welcome.


human_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -83,25 +86,26 @@
...
+        i)"""
 
 
 # Since this experiment contains both RNA-seq and smRNA-seq samples which are in different ranges so we will drop smRNA samples so that samples are within the same range. The analysis identifying these two subsets of samples can be found in this [notebook](../explore_data/0_explore_input_data.ipynb)
 
 # In[5]:
 
 
-# Remove subset of samples
+"""# Remove subset of samples
```
jjc2718, line 108: Do these commented out sections still need to be in this notebook? If you want to remove them but refer back to them in the future if you need them again, you can just remove them - they'll still be in your commit history.

Generally I prefer to just remove things if they're no longer needed rather than commenting them out, since it can be hard to remember in the future why you commented something out or under what conditions you should uncomment it.

If this section of code should run only under certain conditions, it would be best to wrap it in a conditional, or at the very least add a comment describing when it should run.

pseudomonas_analysis/nbconverted/Viz_template_experiment.py
```diff
@@ -0,0 +1,142 @@
...
+    legend_key=element_rect(fill='white', colour='white'), 
+    legend_title=element_text(family='sans-serif', size=15),
+    legend_text=element_text(family='sans-serif', size=12),
+    plot_title=element_text(family='sans-serif', size=15),
+    axis_text=element_text(family='sans-serif', size=12),
+    axis_title=element_text(family='sans-serif', size=15)
+    )
+fig += guides(colour=guide_legend(override_aes={'alpha': 1}))
+
+print(fig)
```
jjc2718, line 141: Were you expecting to see a clear separation between the treated and untreated samples? Does this vary with different random seeds or different UMAP parameters?

README.md
```diff
@@ -53,7 +53,22 @@ pip install -e .
 ```
 
 ## How to run
-The directory paths and parameters will need to be updated in the `config.tsv` file. 
+
+**Steps:**
```
jjc2718, line 57: Maybe "Steps to identify generic DE genes on new data" or something like that would be a clearer heading?

generic_expression_patterns_modules/process.py
```diff
@@ -203,3 +232,92 @@ def generate_summary_table(
...
+        dataframe containing your rank and reference rank per gene
+    reference_rank_col: str
+        column header containing the reference rank
+
+    Returns
+    -------
+    The same merged_gene_ranks dataframe with reference ranks re-scaled
+    """
+    # Scale published ranking to our range
+    max_rank = max(merged_gene_ranks_df["Rank (simulated)"])
```
jjc2718, line 317: Can you use [scikit-learn's MinMaxScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html) for this? It seems like that would be more general.

You'd need to test whether that gives the same results as what you're doing, though - I'm not sure off the top of my head.

pseudomonas_analysis/nbconverted/1_process_pseudomonas_data.py
```diff
@@ -0,0 +1,168 @@
+
+# coding: utf-8
+
+# # Process recount2 data
```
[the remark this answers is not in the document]
ajlee21: correct, copy-paste error!

pseudomonas_analysis/nbconverted/1_process_pseudomonas_data.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Read compendium
+original_compendium = pd.read_csv(original_compendium_file,
+                                  header=0,
+                                  index_col=0,
+                                  sep="\t")
+
+if original_compendium.shape != (950, 5549):
+    original_compendium = original_compendium.T
```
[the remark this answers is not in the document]
ajlee21: good point, the more checks the better:)

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,337 @@
+
+# coding: utf-8
+
+# # Identify generic genes and pathways
+# 
+# This notebooke performs the following steps to identify generic genes
+# 1. Simulates N gene expression experiments using [ponyo](https://github.com/ajlee21/ponyo)
+# 2. Perform DE analysis to get association statistics for each gene
+# 
+# In this case the DE analysis is based on the experimental design of the template experiment, described in the previous [notebook](1_process_recount2_data.ipynb). The template experiment is [SRP012656](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE37764), which contains primary non-small cell lung adenocarcinoma tumors and adjacent normal tissues of 6 never-smoker Korean female patients. So the DE analysis is comparing tumor vs normal in this case.
```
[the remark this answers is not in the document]
ajlee21: Yes, this is another copy error

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,337 @@
...
+                        kind='hex',
+                        marginal_kws={'color':'white'})
+    fig.set_axis_labels("Our preliminary method", "Reference Name", fontsize=14)
+
+    fig.savefig(fig_file,
+                format='svg',
+                bbox_inches="tight",
+                transparent=True,
+                pad_inches=0,
+                dpi=300,)
```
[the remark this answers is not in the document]
ajlee21: In this case, there is nothing plotted because we don't have a reference set of genes that we are comparing against. I kept it here to be consistent with the recount2 analysis so if a user tries to run their own analysis and make a copy of the notebook they didn't have to change cells too much...not sure what the best tactic is for this.

pseudomonas_analysis/nbconverted/Viz_template_experiment.py
ajlee21: I will add a comment here. This visualization was meant to give me a rough sense for the template experiment and the strength of the differential signal between the treated vs untreated samples.

human_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: These are cells that were commented out just so that I could test the other cells. In this case the entire notebook should be run in its entirety and not pieces so I'll uncomment them all

generic_expression_patterns_modules/process.py
ajlee21: Looks like this is doing more of a z-scoring, which isn't what we want in this case

jjc2718: I think it should be a linear scaling to a range with a particular min and max value (the scikit-learn version of z-scoring is StandardScaler). It implements [this formula](https://math.stackexchange.com/a/43699), which I think is the general version of what you're doing.

Anyway, just something to think about for the future. If what you're doing is working fine, no reason to change it now.

ajlee21: AH...you're correct! I misread what was happening. I will definitely change this in the next PR because it will make the pipeline more robust

Issue #8, opened 2020-07-17
Simulated experiments are different between runs


ajlee21:
Though I am setting a numpy.random seed at the top of the simulation function file, the simulated experiments appear to vary between runs. I am assuming this variance based on the confidence interval not matching between runs. 

Not sure where the uncertainty is coming from here.

ajlee21:
Assertion statements to compare simulated experiments generated between different runs passes on local machine (both in jupyter notebooks and using pytest locally). However these assertions fail on github actions. These assertion cells are commented out for now.

PR #9, opened 2020-07-17
Add pytests

ajlee21:
The main change in this PR is to add test files and notebooks to run using github actions

Additional minor changes include:
1. Updates `process.py` module to use sklearn minmax scaler to make the pipeline more flexible
2. Updated `environment.yml` file to install ponyo using pip

In a previous PR, there was the suggestion to revise the config file so that the user did not need to manually add the paths. I have create a `paths_test.py` file to test out using both a paths and config file per notebook, but this lead to needing to change the notebook so I am not using this paths file currently. I'm not sure the best way to resolve this. I plan to have different config files per analysis and don't want to make the user manually copy and paste paths if possible.


README.md
```diff
@@ -14,28 +14,24 @@
...
-**Problem**: 
-* This method required a large manual curation effort
-* If you want to perform a new DE analysis in a different biological **context** (i.e. different organism, tissue, media) then you might not have the curated data available. Switch contexts will require a lot of manual effort. 
-* Similarly, using a different statistical method will require re-curation effort
+**Importance:**
+Why is it important to identify generic genes and pathways?
 
-**Question**: Can we use our gene expression simulator, [ponyo](https://github.com/greenelab/ponyo), automatically generate null experiments for different contexts in order to overcome this manual effort?
+**Challenge**: 
+Identification of generic genes and pathways required a large manual curation effort. If you want to perform a new DE analysis in a different biological **context** (i.e. different organism, tissue, media) then you might not have the curated data available. Switch contexts will require a lot of manual effort. Similarly, using a different statistical method will require re-curation effort
```
jjc2718, line 21:
```suggestion
Identification of generic genes and pathways required a large manual curation effort. If you want to perform a new DE analysis in a different biological **context** (i.e. different organism, tissue, media) then you might not have the curated data available. Switching contexts will require a lot of manual effort. Similarly, using a different statistical method will require re-curation effort
```

generic_expression_patterns_modules/paths_test.py
```diff
@@ -0,0 +1,26 @@
+"""Path definitions"""
+from pathlib import Path
+
+# Path to this repository
+PROJECT_DIR = Path(__file__).parents[1]
+ANALYSIS_DIR = PROJECT_DIR / "sra_experiment"
```
jjc2718, line 6: I didn't know you could do this with `pathlib` and the division operator. Neat!

README.md
jjc2718, line 21:
```suggestion
Identification of generic genes and pathways required a large manual curation effort. If you want to perform a new DE analysis in a different biological **context** (i.e. different organism, tissue, media) then you might not have the curated data available. Switching contexts will require a lot of manual effort. Similarly, using a different statistical method will require re-curation effort
```

human_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -242,6 +242,17 @@
     gene_summary_file, sep='\t')
 
 
+# In[18]:
+
+
+# Test reproducibility
```
jjc2718, line 248: Do you want to keep this here, or was this just a one-time test? Or should this be in the version of the notebook in the `tests` directory?

human_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -282,6 +293,7 @@
                                              shared_gene_rank_scaled_df,
                                              1000)
     print(r, p, ci_high, ci_low)
+    # 0.2397233887765045 3.960267987659121e-223 0.22519096058775306 0.25479874632262184
```
jjc2718, line 296: This seems like something you might want to put in your notebook in `tests/2_identify_generic_genes_pathways.ipynb`, doesn't it? You could have an assert statement making sure the correlation is what you expect it to be.

tests/2_identify_generic_genes_pathways.ipynb
```diff
@@ -0,0 +1,740 @@
+{
```
jjc2718, line 1: Following up on my comment in `human_analysis/2_identify_generic_genes_pathways.ipynb`, I guess I'm a bit confused on what the purpose of this test is. Are you trying to make sure this notebook runs without errors? If so, does that mean you'll need to remember to update these notebooks in the `tests` directory whenever you change the other notebooks?

I'm not necessarily opposed to this, just trying to get a handle on what your goal is with testing here. What you're doing is a bit different than the typical unit testing I see in other people's repos, so I just want to get your perspective on it.

human_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: This was a one time test that really should be in the test directory, you're right

human_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Yes, good point. I have added an assertion here

tests/2_identify_generic_genes_pathways.ipynb
ajlee21: Thank you for thinking about this. My goal was to make sure that all the analysis runs without errors, so testing all the modules run at least. I tried to add assertion statements where I could (i.e. checking the VAE output and correlation outputs were the same). 

I have tried to add additional checks to test:
- that simulated experiments are reproducible
- that the ranks are reproducible

You're right that with the current setup I would need to update the test notebook if the analysis notebooks change to make sure they're up to date. Any advice you have to make these tests better is very welcome

jjc2718: Makes sense. Maybe you could have a step in the workflow in `tests.yaml` to copy over the script from `human_analysis/2_identify_generic_genes_pathways.ipynb` to the `tests` directory? Then you wouldn't have the extra tests that you added just now, though, so you'd have to make a separate notebook to run those.

I don't think this needs to be done now - maybe you could create an issue for this and come back to it in the future?

ajlee21:
The added assertion statements that pass locally seem to error out using github actions. For some reason the simulated experiments generated locally are consistent but fail to be `isclose` on github actions for certain genes and samples. I have updated the issue: https://github.com/greenelab/generic-expression-patterns/issues/8 

Since the runs seem to be consistent locally at least, I will leave this for now

tests/2_identify_generic_genes_pathways.ipynb
ajlee21: Thank you for thinking about this with me. So that analysis script is mainly calling other functions/modules and would be nice to automate but I'm ok manually porting things for now. I guess I'm not clear on how to write a script to port things over..

PR #17, opened 2020-09-18
Add functions to create compendium file for all recount2 data

dongbohu:
This PR includes:
- a few new functions in Python and R to download all recount2 gene counts data and compile them into a single compendium file in `tsv` format. 
- coding style changes in R scripts. (Some of them were done automatically by my editor.) The coding style guide that I am using is this one: https://style.tidyverse.org/index.html, which seems to be a popular one.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ channels:
 - bioconda
 dependencies:
 - r::r=3.6.0
-- r::r-devtools 
+- r::r-devtools
 - bioconda::bioconductor-limma
 - bioconda::bioconductor-deseq2
 - bioconda::bioconductor-recount
@@ -22,6 +22,7 @@ dependencies:
 - conda-forge::gxx_linux-64
 - conda-forge::rpy2=2.9.4
 - conda-forge::libiconv=1.15
+- conda-forge::r-biocmanager
 - pip=19.2.1
 - pip:
   - matplotlib==3.0.0
--- a/environment.yml
+++ b/environment.yml
@@ -7,6 +7,7 @@ channels:
 dependencies:
 - r::r=3.6.0
 - r::r-devtools
+- r::r-stringi
 - bioconda::bioconductor-limma
 - bioconda::bioconductor-deseq2
 - bioconda::bioconductor-recount
--- a/generic_expression_patterns_modules/download_recount2_data.R
+++ b/generic_expression_patterns_modules/download_recount2_data.R
@@ -1,82 +1,77 @@
 ## Run this once to setup environment
 ## Used R 3.6.3
-#if (!requireNamespace("BiocManager", quietly = TRUE))
-#  install.packages("BiocManager")
 
-#BiocManager::install("recount")
+library('recount')
 
 ## About
 # recount2 contains over 70,000 uniformly processed human RNA-seq samples spanning
 # TCGA, SRA and GTEx
-# library('recount')
 
 set.seed(123)
 
-get_recount2_template_experiment <- function(project_id,
-local_dir){
-
+get_recount2_template_experiment <- function(project_id, local_dir) {
   # This function downloads the gene expression data associated
   # with the input project_id and saves it in the local_dir
+
+  # Reset working directory to `local_dir`
+  setwd(local_dir)
 
   ## Data
 
   # Get data associated with project ids
-  # Download the RangedSummarizedExperiment object.
-  # The RangedSummarizedExperiment object for the counts summarized 
-  # at the gene level using the Gencode v25 (GRCh38.p7, CHR) annotation as provided by Gencode.
+  # Download the RSE (RangedSummarizedExperiment) object.
+  # The RSE object for the counts summarized at the gene level using the
+  # Gencode v25 (GRCh38.p7, CHR) annotation as provided by Gencode.
   if (!file.exists(file.path(project_id, 'rse_gene.Rdata'))) {
     download_study(project_id)
-  } 
-  load(file.path(project_id, 'rse_gene.Rdata'), verbose = TRUE)
+  }
+  load(file.path(project_id, 'rse_gene.Rdata'), verbose=TRUE)
 
-  # Counts are raw read counts from the sequencing run
   # Counts are the number of reads that map that each gene
-
-  # Rename counts for storage
-  assign(paste0("rse_gene_", project_id), rse_gene)
 
   rse_gene_scaled <- scale_counts(rse_gene)
   rse_gene_counts <- assays(rse_gene_scaled)$counts
   data_counts <- t(rse_gene_counts)
 
   ## Save counts matrix to file
-  write.table(data_counts,
-            paste(local_dir,'/recount2_template_data.tsv',sep=""),
-            sep='\t',
-            row.names=TRUE,
-            col.names=TRUE)
-  
-  ## Delete it if you don't need it anymore
-  unlink(project_id, recursive = TRUE)
+  write.table(
+    data_counts,
+    paste(local_dir, '/recount2_template_data.tsv', sep = ""),
+    sep = '\t',
+    row.names = TRUE,
+    col.names = NA
+  )
 
 }
 
 
-get_recount2_compendium <- function(template_project_id,
-num_experiments,
-local_dir,
-base_dir){
-
-  # This function downloads 50 random experiments, including the 
+get_recount2_subset_compendium <- function(template_project_id,
+                                           num_experiments,
+                                           local_dir,
+                                           base_dir) {
+  # This function downloads 50 random experiments, including the
   # the experiment associated with the input project_id.
   # These experiments makeup a compendium of gene expression experiments
   # that are saved in the local_dir
 
   # Download metadata file
   metadata <- all_metadata()
-  write.table(metadata, 
-              paste(base_dir, '/human_analysis/data/metadata/recount2_metadata.tsv', sep=""),
-              sep='\t', 
-              row.names=FALSE)
+  write.table(
+    metadata,
+    paste(base_dir, '/human_analysis/data/metadata/recount2_metadata.tsv', sep=""),
+    sep = '\t',
+    row.names = FALSE
+  )
 
   # Format of the metadata
-  # Based on definitions from NCBI and blog post: 
+  # Based on definitions from NCBI and blog post:
   # https://www.ncbi.nlm.nih.gov/books/NBK56913/#search.what_do_the_different_sra_accessi
   # https://www.ccdatalab.org/blog/2019/3/29/gene-expression-repos-explained
-  # Project: A sequencing study (i.e. NCBI sequencing study).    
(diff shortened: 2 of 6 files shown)


environment.yml
```diff
@@ -22,6 +22,7 @@ dependencies:
 - conda-forge::gxx_linux-64
 - conda-forge::rpy2=2.9.4
 - conda-forge::libiconv=1.15
+- conda-forge::r-biocmanager
```
dongbohu, line 25: I am adding this package so we don't have to worry about whether `BiocManager` is installed or not in R sessions.

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -1,82 +1,77 @@
 ## Run this once to setup environment
 ## Used R 3.6.3
-#if (!requireNamespace("BiocManager", quietly = TRUE))
-#  install.packages("BiocManager")
 
-#BiocManager::install("recount")
```
dongbohu, line 6: I deleted this line because `recount` has been installed in conda env.

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -1,82 +1,77 @@
...
 }
 
 
-get_recount2_compendium <- function(template_project_id,
-num_experiments,
-local_dir,
-base_dir){
-
-  # This function downloads 50 random experiments, including the 
+get_recount2_subset_compendium <- function(template_project_id,
```
dongbohu, line 48: I renamed `get_recount2_compendium` to `get_recount2_subset_compendium`, maybe we don't need this function any more?

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -123,9 +119,62 @@ base_dir){
...
+    paste(local_dir,'/recount2_compendium_data.tsv',sep = ""),
+    sep = '\t',
+    row.names = TRUE,
+    col.names = NA
+  )
+}
+
+# This function downloads all recount2 `rse_gene.Rdata` files and generates a
+# transposed data count file (`t_data_counts.tsv`) based on each `rse_gene.Rdata`.
+function download_all_recount2(metadata_dir, download_dir) {
```
dongbohu, line 133: This function is based on the original `get_recount2_compendium`. The biggest difference is that it saves the transposed `data_counts` for each project on disk. I have a new function in `process.py` to read the individual `t_data_counts.tsv` on disk and create the final compendium file. The main reason to do this is because the line:
```
data_counts_all <- cbind(data_counts_all, assays(scale_counts(rse_gene))$counts)
``` 
required a huge amount of memory when `data_counts_all` gets big.

generic_expression_patterns_modules/process.py
```diff
@@ -416,11 +416,41 @@ def compare_and_reorder_samples(expression_file, metadata_file):
...
     else:
         # Convert gene expression ordering to be the same as
         # metadata sample ordering
         print("sample ids don't match, going to re-order gene expression samples")
         expression_data = expression_data.reindex(metadata_sample_ids)
         expression_data.to_csv(expression_file, sep="\t")
-        return
+
+
+def create_all_recount2_compendium(download_dir, output_filename):
```
dongbohu, line 427: This is the function that reads each data_counts file and combine them together into compendium file.

dongbohu:
@ajlee21 I added some comments in this PR to explain why I made some changes. Please let me know if you need more clarification.

generic_expression_patterns_modules/process.py
dongbohu: I need to figure out why the checks failed.

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: By the way, most of the changes I made on this function are trivial coding styles, such as `sep=""` to `sep = ""`.

environment.yml
ajlee21: Does this mean I will not need to have these cells in my analysis notebooks:
```
#if (!requireNamespace("BiocManager", quietly = TRUE))
#    install.packages("BiocManager")
#BiocManager::install("derfinder")
#BiocManager::install("recount")
```

That would be nifty

generic_expression_patterns_modules/download_recount2_data.R
ajlee21: I thought I tried tried to remove this and it didn't work. But if you have tried removing it and its not needed that's great. It was a while ago so I might be misrembering

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -1,82 +1,77 @@
...
 
   ## Save counts matrix to file
-  write.table(data_counts,
-            paste(local_dir,'/recount2_template_data.tsv',sep=""),
-            sep='\t',
-            row.names=TRUE,
-            col.names=TRUE)
-  
-  ## Delete it if you don't need it anymore
-  unlink(project_id, recursive = TRUE)
```
ajlee21, line 50: I added this line to remove the intermediate downloaded files, is this not needed because you deleted the `assign` call?

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -123,9 +119,62 @@ base_dir){
...
+    paste(local_dir,'/recount2_compendium_data.tsv',sep = ""),
+    sep = '\t',
+    row.names = TRUE,
+    col.names = NA
+  )
+}
+
+# This function downloads all recount2 `rse_gene.Rdata` files and generates a
+# transposed data count file (`t_data_counts.tsv`) based on each `rse_gene.Rdata`.
+download__all_recount2 <- function (metadata_dir, download_dir) {
```
ajlee21, line 133: Do you recall how long the download took? I wonder if it'd be helpful to add in a comment so that the person knows how long they can expect to wait

generic_expression_patterns_modules/process.py
```diff
@@ -416,11 +416,41 @@ def compare_and_reorder_samples(expression_file, metadata_file):
...
+    The first row in each `t_data_counts.tsv` is a header line that includes
+    column names, so only the header in the first `t_data_counts.tsv` is copied
+    to the output file.
+
+    Arguments:
+      - download_dir: the dirname that hosts all downloaded projects data
+      - output_filename: the filename of the output single compendium data
+    """
+
+    from glob import glob
```
ajlee21, line 440: Minor - i tend to put all my imports at the top of the file.

generic_expression_patterns_modules/process.py
ajlee21: Just curious, was there a reason to split up the downloading the concatentation step? You can probably see that when I coded it, I downloaded a file, concatenated it to a dataframe and then deleted the downloaded file. But I'm not sure how efficient that was and I think you have more experience with large data, so I was just curious to learn

generic_expression_patterns_modules/download_recount2_data.R
ajlee21: Thank you for renaming this. I don't think we should need this now that you can train on all of recount2, but I think its fine to leave it for now. Perhaps this'll be an option that we decide we want in the future...

ajlee21 (approved):
Code all looks good to me, just had a few questions places.

What are the checks that are failing? Can I help with this?

environment.yml
dongbohu: These two lines
```
#if (!requireNamespace("BiocManager", quietly = TRUE))
#    install.packages("BiocManager")
```
won't be needed.

For `derfinder` (line 3), we still need to install it in conda's `environment.yml` file (if that package is available in conda). I will take care of that in later PR(s). 

For `recount` (line 4), see my comments below.

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: After creating a brand new conda env, I tested `library("recount")` command in a R session and got the following error:
```
Error: package or namespace load failed for ‘recount’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/home/ubuntu/miniconda3/envs/dhu_test/lib/R/library/stringi/libs/stringi.so':
  libicui18n.so.64: cannot open shared object file: No such file or directory
```
More details on this error can be found here:
https://github.com/conda-forge/r-stringi-feedstock/issues/20
To make a long story short, we need to explicitly install `r-stringi`  in `r` channel in `environment.yml`. I already tested it in another conda env. I will add the line in `environment.yml`.

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: Sure, let's keep this function for now.

generic_expression_patterns_modules/process.py
dongbohu: Good question. The main reason why I separated the downloading and concatenation is because R is usually slower when dealing with a lot of disk I/O, which is required in concatenation process. 

Another reason is because the downloading may take a while, so I prefer to let the downloading finish first. Originally I started the downloading on my local laptop that used WiFi. After a few hours I gave up because WiFi in my house is both slow and unstable. So I logged into my lab linux box and downloaded the recount2 data via ethernet. It was much faster (took less than 30 minutes).

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: I wanted to keep the downloaded file (at least temporally) so that if something went wrong, we don't have to download it again. (It is easy to delete them when the whole data process is done anyway.)

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: I thought of adding some comments on how long the downloading process may take, but it largely depends on the user's network speed (as I mentioned in one of my earlier comments). Maybe I should add a comment that sounds vague, such as `The downloading process may take 30 minutes to a few hours, depending on your network speed.`

dongbohu:
@ajlee21 I make some changes based on your comments. I don't understand why the checks failed on Friday, because the new code should be completely compatible with the older version. Let's whether the checks will pass with my latest commit.

dongbohu:
@ajlee21 I think I know why the checks failed. It is because of this line `BiocManager::install(...)` in notebook. The namespace `BiocManager` is not loaded into R session. Let me fix it after the opening scrum.

generic_expression_patterns_modules/download_recount2_data.R
ajlee21: I see, in that case I'll defer to you. I am not as expert here, so I just assumed the download would be a static amount of time, which in that case I think the time estimate would be helpful but perhaps not given what you're saying

generic_expression_patterns_modules/process.py
ajlee21: Ah..this makes sense. Thank you for the explanation

PR #18, opened 2020-09-22
Add functions to process recount2 template and compendium data

dongbohu:
This PR adds a few functions to process recount2 template and compendium data. Most logics are copied from
https://github.com/greenelab/generic-expression-patterns/blob/master/human_analysis/1_process_recount2_data.ipynb

Here is an example to run these functions:
```python
# input files (assume they already exist)
raw_filename = "raw_template.tsv"
gene_id_filename = "ensembl_hgnc_mapping.tsv"
DE_prior_filename = "https://raw.githubusercontent.com/maggiecrow/DEprior/master/DE_Prior.txt"
sample_id_metadata_filename = "sample_id_meta.tsv"

# User's own gene mapping, which overrides all other mappings
manual_mapping = {
    "ENSG00000187510.7": "PLEKHG7",    # not necessary (available in latest gene_id_mapping)
    "ENSG00000230417.11": "LINC00595", # not necessary (1st of duplicates)
    "ENSG00000276085.1": "CCL3L1",     # not necessary (1st of duplicates)
    "ENSG00000255374.3": "TAS2R45",    # NEEDED        (2nd of duplicates)
}

# Output files
mapped_filename = "mapped_temp.tsv"
processed_filename = "processed_temp.tsv"

# Process recount2 template
process.process_raw_template(
    raw_filename,
    gene_id_filename,
    manual_mapping,
    DE_prior_filename,
    mapped_filename,
    sample_id_metadata_filename,
    processed_filename
)
```

### DIFF
--- a/generic_expression_patterns_modules/process.py
+++ b/generic_expression_patterns_modules/process.py
@@ -433,9 +433,12 @@ def create_all_recount2_compendium(download_dir, output_filename):
     column names, so only the header in the first `t_data_counts.tsv` is copied
     to the output file.
 
-    Arguments:
-      - download_dir: the dirname that hosts all downloaded projects data
-      - output_filename: the filename of the output single compendium data
+    Arguments
+    ---------
+    download_dir: str
+        the dirname that hosts all downloaded projects data
+    output_filename: str
+        the filename of the output single compendium data
     """
 
     data_counts_filenames = glob(f"{download_dir}/*/t_data_counts.tsv")
@@ -457,7 +460,14 @@ def create_all_recount2_compendium(download_dir, output_filename):
 
 
 def get_published_generic_genes(filename):
-    """Get generic genes based on input filename, which could be a URL."""
+    """
+    Get generic genes based on input filename, which could be a URL.
+
+    Arguments
+    ---------
+    filename: str
+        name of the file that includes published generic genes
+    """
 
     df = pd.read_csv(filename, header=0, sep="\t")
     published_generic_genes = list(df['Gene_Name'])
@@ -468,9 +478,16 @@ def get_merged_gene_id_mapping(gene_id_filename, raw_ensembl_genes):
     """
     Merge genes in input gene_id file with the raw ensembl gene IDs.
 
-    Arguments:
-    - gene_id_filename: filename of input gene IDs;
-    - raw_ensembl_genes: list of strings (ensembl gene IDs)
+    Arguments
+    ---------
+    gene_id_filename: str
+        filename of input gene IDs;
+    raw_ensembl_genes: list
+        list of strings (ensembl gene IDs)
+
+    Returns
+    -------
+    Mapping between ensembl ids and hgnc symbols
     """
 
     original_gene_id_mapping = pd.read_csv(
@@ -510,13 +527,19 @@ def get_renamed_columns(
     """
     Find the new column names and corresponding column indexes.
 
-    Arguments:
-    - raw_ensembl_ids: list of strings (ensembl gene IDs), which are columns
-                       names in raw recount2 data file;
-    - merged_gene_id_mapping: DataFrame of merged gene ID mapping;
-    - manual_mapping: dict of manual mapping (key: ensembl_id, value: gene symbol)
+    Arguments
+    ---------
+    raw_ensembl_ids:
+        list of strings (ensembl gene IDs), which are columns names in
+        raw recount2 data file;
+    merged_gene_id_mapping: DataFrame
+        merged gene ID mapping;
+    manual_mapping: dict
+        dict of manual mapping (key: ensembl_id, value: gene symbol)
 
-    Returns a tuple that includes two entries. The first entry is a list
+    Returns
+    -------
+    A tuple that includes two entries. The first entry is a list
     of hgnc gene symbols (which will be the new column names in remapped
     recount2 data file; The second entry is a dict whose keys are hgnc gene
     symbols and values are lists of the corresponding indexes of columns in
@@ -628,7 +651,7 @@ def process_raw_template(
     template data on disk.
     """
 
-    # Create mapped recount2 template data file
+    # Write the intermediate mapped recount2 template data file on disk
     map_recount2_data(
         raw_filename,
         gene_id_filename,
@@ -637,17 +660,20 @@ def process_raw_template(
         mapped_filename
     )
 
-    # Drop sample rows
+    sample_ids_to_drop = set()
     if os.path.exists(sample_id_metadata_filename):
         # Read in metadata and get samples to be dropped:
         metadata = pd.read_csv(
             sample_id_metadata_filename, sep='\t', header=0, index_col=0
         )
-        sample_ids_to_drop = list(metadata[metadata["processing"] == "drop"].index)
+        sample_ids_to_drop = set(metadata[metadata["processing"] == "drop"].index)
 
-        template_df = pd.read_csv(mapped_filename, sep='\t', header=0, index_col=0)
-        template_df = template_df.drop(sample_ids_to_drop)
-        template_df.to_csv(processed_filename, sep='\t')
+    # Write the processed recount2 template output file on disk
+    with open(mapped_filename) as ifh, open(processed_filename, "w") as ofh:
+        for idx, line in enumerate(ifh):
+            sample_id = line.split('\t')[0]
+            if idx == 0 or sample_id not in sample_ids_to_drop:
+                ofh.write(line)
 
 
 def normalize_compendium(mapped_filename, normalized_filename):
@@ -671,8 +697,8 @@ def normalize_compendium(mapped_filename, normalized_filename):
     normalized_compendium = scaler.fit_transform(mapped_compendium_df)
     normalized_compendium_df = pd.DataFrame(
         normalized_compendium,
-        columns=processed_compendium.columns,
-        index=processed_compendium.index
+        columns=mapped_compendium_df.columns,
+        index=mapped_compendium_df.index
     )
 
     # Save normalized data on disk: ~17.5 minutes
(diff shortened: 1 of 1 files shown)


dongbohu:
@ajlee21: I don't know how much you like or hate the idea of moving the logic in the notebook to standalone functions. It seems a good idea from a developer's point of view, but it may not from a user's point of view. Let me what you think.

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+        how='outer'
+    )
+
+    # Set `ensembl_version` column as the index
+    merged_gene_id_mapping.set_index('ensembl_version', inplace=True)
+
+    return merged_gene_id_mapping
+
+
+def get_renamed_columns(
```
dongbohu, line 504: This function probably will give most trouble to understand. The idea is that if we read the original compendium data file is too big to load into the memory. To map the genes in it, we only need its first row (header line). After we figure out how the columns can be mapped from ensembl names to hgnc gene symbols, we will read it line by line and remove unnecessary columns.

dongbohu:
In this PR, I am referring to:
- the original template or compendium data (generated directly from the downloaded recount2 data) as `raw` file
- the one whose column names have been mapped from ensembl IDs to gene symbols as `mapped` file
- the final template output file as `processed` (some sample rows may have dropped)
- the final compendium output file as `normalized`

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
 
                 file_content = ifh.read()
                 ofh.write(file_content)
+
+
+def get_published_generic_genes(filename):
+    """Get generic genes based on input filename, which could be a URL."""
```
ajlee21, line 460:
```suggestion
    """
    Get generic genes based on input filename, which could be a URL.
    """
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    df = pd.read_csv(filename, header=0, sep="\t")
+    published_generic_genes = list(df['Gene_Name'])
+    return published_generic_genes
+
+
+def get_merged_gene_id_mapping(gene_id_filename, raw_ensembl_genes):
+    """
+    Merge genes in input gene_id file with the raw ensembl gene IDs.
+
+    Arguments:
```
ajlee21, line 471:
```suggestion
    Arguments
    ------------
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    - gene_id_filename: filename of input gene IDs;
```
ajlee21, line 472:
```suggestion
    gene_id_filename: str
        filename of input gene IDs
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    - raw_ensembl_genes: list of strings (ensembl gene IDs)
```
ajlee21, line 473:
```suggestion
    - raw_ensembl_genes: list of strings (ensembl gene IDs)
```
```suggestion
    raw_ensembl_genes: list
        list of strings (ensembl gene IDs)
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    """
```
ajlee21, line 474:
```suggestion
    Returns
    ---------
    Mapping between ensembl ids and hgnc symbols
    """
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+        raw_ensembl_ids,
+        merged_gene_id_mapping,
+        manual_mapping,
+        DE_prior_filename
+):
+    """
+    Find the new column names and corresponding column indexes.
+
+    Arguments:
+    - raw_ensembl_ids: list of strings (ensembl gene IDs), which are columns
```
ajlee21, line 514:
```suggestion
    raw_ensembl_ids: list
        list of strings (ensembl gene IDs), which are columns
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+def get_renamed_columns(
+        raw_ensembl_ids,
+        merged_gene_id_mapping,
+        manual_mapping,
+        DE_prior_filename
+):
+    """
+    Find the new column names and corresponding column indexes.
+
+    Arguments:
```
ajlee21, line 513:
```suggestion
    Arguments
    ------------
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    - raw_ensembl_ids: list of strings (ensembl gene IDs), which are columns
+                       names in raw recount2 data file;
+    - merged_gene_id_mapping: DataFrame of merged gene ID mapping;
```
ajlee21, line 516:
```suggestion
    merged_gene_id_mapping: dataframe
        DataFrame of merged gene ID mapping
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    - manual_mapping: dict of manual mapping (key: ensembl_id, value: gene symbol)
```
ajlee21, line 517:
```suggestion
    manual_mapping: dict
        dict of manual mapping (key: ensembl_id, value: gene symbol)
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+            hgnc_to_cols[hgnc_symbol] = [col_idx]
+
+    our_gene_ids_hgnc = list(hgnc_to_cols.keys())
+
+    published_generic_genes = get_published_generic_genes(DE_prior_filename)
+    shared_genes_hgnc = list(
+        set(our_gene_ids_hgnc).intersection(published_generic_genes)
+    )
+
+    # In Python, the order of elements in a list that is converted from a set
```
ajlee21, line 558:
```suggestion
    # In Python, the order of elements in a list that is converted from a set,
```

generic_expression_patterns_modules/process.py
```diff
@@ -454,3 +454,256 @@ def create_all_recount2_compendium(download_dir, output_filename):
...
+    # is non-deterministic, so it is sorted here to have reproducible result.
```
ajlee21, line 559:
```suggestion
    # which is non-deterministic, so it is sorted here to have reproducible result.
```

ajlee21 (approved):
This all looks SO great! Thank you so much for cleaning up this repo!

My changes are just to keep a consistent style with the other modules.

I really like the idea of moving modular size pieces of the the notebook (https://github.com/ajlee21/generic-expression-patterns/blob/master/human_analysis/1_process_recount2_data.ipynb) into functions. I think having the notebook calling these separate functions will be a lot cleaner easier to understand and run.

dongbohu:
@ajlee21: I noticed aoo.1 minor issue in my new approach to process the template or compendium file. Suppose we have a DataFrame in Pandas like this:
```
    foo foo bar
r1   .   .    .
r2   .   .    .
```
then when it is saved by `pandas.to_csv()` function, the column names that are not unique (`foo`) will be renamed to `foo.1`, 
so the output tsv file will look like:
```
    foo foo.1 bar
r1   .   .    .
r2   .   .    .
```
but my approach doesn't use Pandas, so the identical column names are still identical in output tsv file. Does it affect any downstream analysis? Even if it doesn't, do you prefer changing the second `foo` column name into `foo.1`?

ajlee21:
Based on my recollection, there were not many of these cases of duplicates and there is a step [here](https://github.com/greenelab/generic-expression-patterns/blob/1727123cfcc8a2e9adfc9112d62710307212e174/human_analysis/nbconverted/1_process_recount2_data.py#L231) that filters the genes to only include those that are shared between our dataset and the published dataset, so I very much doubt that these will change the downstream results we see.

I can see two options for this:
1. Check to see if the duplicate happens AFTER the mapping function is applied. In other words, is there an ensembl id that maps to multiple hgnc symbols and this is causing the duplicate)? If this is the case I tried to manuallly adjust for these [here](https://github.com/greenelab/generic-expression-patterns/blob/1727123cfcc8a2e9adfc9112d62710307212e174/generic_expression_patterns_modules/process.py#L71)

If the duplicate exists in the dataset BEFORE the mapping is applied, then I would look to see if the column values for these duplicate gene ids are also the same. If they are then it should matter which one to take. If they are different then I would say to drop them both because its unclear how to interpret these.

2. If it would be too difficult/confusing for you to do the checks listed above, then you can make an issue and I'd be happy to look into it because, like i said, I don't think this should change the result much.

generic_expression_patterns_modules/process.py
dongbohu: Actually, what I meant here is "the order (of ...) is non-deterministic, ...", so `which` is not needed.

dongbohu:
@ajlee21: Never mind my question above. I figured out why. It's because of `read_csv` function in pandas, not `to_csv`. The duplicates happened after the mapping, and my code already took care of it (based on your logic).

PR #19, opened 2020-09-23
Revise human_analysis/1_process_recount2_data.ipynb

dongbohu:
This PR includes:
- changes on the first ipython notebook for human analysis
- fine tuning of a few functions in the module
- a few more configuration parameters used in recount2 data processing


dongbohu:
@ajlee21: This PR is more like a work-in-progress. I need to run the ipython notebook to make sure that it is working fine. Feel free to run it by yourself too.

human_analysis/1_process_recount2_data.ipynb
```diff
@@ -48,12 +50,10 @@
     "import pandas as pd\n",
     "import numpy as np\n",
     "import rpy2\n",
```
dongbohu, line 52: Question: Do we need to import this module explicitly?

human_analysis/1_process_recount2_data.ipynb
```diff
@@ -48,12 +50,10 @@
...
-    "import seaborn as sns\n",
-    "from sklearn import preprocessing\n",
     "import pickle\n",
     "\n",
     "from ponyo import utils, train_vae_modules\n",
-    "from generic_expression_patterns_modules import process, calc\n",
+    "from generic_expression_patterns_modules import process\n",
     "\n",
     "np.random.seed(123)"
```
dongbohu, line 58: Question: Why we need to set the random seed?

human_analysis/1_process_recount2_data.ipynb
```diff
@@ -4718,21 +3059,15 @@
...
     "# Setup directories\n",
     "# Create VAE directories\n",
-    "output_dirs = [os.path.join(base_dir, dataset_name, \"models\"),\n",
-    "               os.path.join(base_dir, dataset_name, \"logs\")]\n",
-    "\n",
-    "# Check if analysis output directory exist otherwise create\n",
-    "for each_dir in output_dirs:\n",
-    "    if os.path.exists(each_dir) == False:\n",
-    "        print('creating new directory: {}'.format(each_dir))\n",
-    "        os.makedirs(each_dir, exist_ok=True)\n",
```
dongbohu, line 4728: This section is not necessary, because `os.makedirs(dirname, exist_ok=True)` will always generate the parent directories if they don't exist yet.

human_analysis/1_process_recount2_data.ipynb
```diff
@@ -4718,21 +3059,15 @@
...
+    "output_dirs = [\n",
+    "    os.path.join(base_dir, dataset_name, \"models\"),\n",
+    "    os.path.join(base_dir, dataset_name, \"logs\")\n",
+    "]\n",
     "\n",
     "# Check if NN architecture directory exist otherwise create\n",
     "for each_dir in output_dirs:\n",
     "    new_dir = os.path.join(each_dir, NN_architecture)\n",
-    "    if os.path.exists(new_dir) == False:\n",
```
dongbohu, line 4733: I removed this `if` block because with `exist_ok=True`, `os.makedirs(...)` will always succeed, no matter the directory that you want to create exists or not.

human_analysis/1_process_recount2_data.ipynb
ajlee21: Tried commenting this `rpy2` module and it doesn't look like you do need this. Thanks for checking

human_analysis/1_process_recount2_data.ipynb
ajlee21: This seed was added to make sure that results of this notebook were reproducible but I don't think I need it here since `ponyo` is has its own settings to ensure reproducible training results and now we're sampling all of the SRA data. So I think this can be removed

ajlee21:
In the process of testing running the notebook. Here are just a few comments so far:
* Would it make sense to add a check if `download_recount2_sra` finished so that this step could be skipped?

Update text:
* "1. Select template experiment" to "1. Select template experiment. This template experiment will be used in the next [notebook](2_identify_generic_genes_pathways.ipynb) to simulate experiments with the same experimental design but testing a different biological process."
* Need to delete space between "[coverage counts]" and "(https://www.bioconductor.org/packages/devel/workflows/vignettes/recountWorkflow/inst/doc/recount-workflow.html)"

* Update markdown cell to be consistent with cell below "Process recount2 template experiment data" to "Map ensembl gene IDs in template experiment data"

ajlee21 (approved):
Looks good. I added a few minor changes after running the notebook. the notebook is SO much cleaner and very easy to read now..thank you!

Just had a few comments above. I think the main one is about the error for the one project id.

PR #23, opened 2020-09-30
Refactoring: human_analysis/2_identify_generic_genes_pathways.ipynb

dongbohu:
This PR refactors some functionalities in `human_analysis/2_identify_generic_genes_pathways.ipynb` into functions, and cleans up some code style issues.


ajlee21 (commented):
This all looks great! Much more concise..thank you for pulling out the modules again.

I'd like to test this notebook once the internet returns.

ajlee21:
Only one issue when testing the notebook runs with the updated env!
I got error: running this cell
```
%%R -i template_DE_stats_filename -i hallmark_DB_filename -i statistic -o template_enriched_pathways

source('../generic_expression_patterns_modules/GSEA_analysis.R')

template_enriched_pathways <- find_enriched_pathways(template_DE_stats_filename, hallmark_DB_filename, statistic)
```
Error message: `Error in file(con, "r") : cannot open the connection`. If I set `setwd("/home/alexandra/Documents/Repos/generic-expression-patterns/generic_expression_patterns_modules")` before I call this cell it works. I guess the working dir needs to be set for all of these R calls.

Some minor changes I found:
1. In the config file: `reference_gene_file` needs to be updated to `reference_gene_filename`
2. I think we need to add the `processed_recount2_template.tsv`, `hallmark_DB.gmt` and `scaler_transform_human.pickle` files to the repo. And these locations need to be added to the Readme

ajlee21 (approved):
Looks like its good to merge once those minor changes are fixed

dongbohu:
@ajlee21: Thank you for your comments. I can't reproduce the error that was generated by `%%R -i template_DE_stats_filename ...` command. Do you think the error was caused by the following line:
```
source('../generic_expression_patterns_modules/GSEA_analysis.R')
```

Good catch on `reference_gene_filename`, I will modify it.

About the three new files (processed_recount2_template.tsv, hallmark_DB.gmt, scaler_transform_human.pickle), where should we put them in the repo?

ajlee21:
Yes I believe its the `source` call that is causing the issue. The same issue happened to me for notebook 1 because the R download errored out so the resetting of the working directory (`setwd(original_wd)`) that you have at the end the script didn't run. I found that setting the working directory with that one line fixed that issue.

I never really delt with setting the working directory prior to this so I'm not sure why all of a sudden there is this issue. 

As for the new data I think
`hallmark_DB.gmt` --> `data/metadata/`
`processed_recount2_template.tsv` --> `data/`
`scaler_transform_human.pickle` --> `data/`

dongbohu:
What I don't understand is that if the `source(...)` line causes error here, it should have caused error earlier too. When you run the second notebook, did you accidentally change the working directory before the `source` line?

dongbohu:
Adding `setwd("/home/alexandra/Documents/Repos/generic-expression-patterns/generic_expression_patterns_modules")` in all R calls is not a good idea, because any user other than you is likely to have a different location of `generic_expression_patterns_modules` directory, so this hard-coded location wouldn't work for them.

dongbohu:
I changed this R session into:
```
%%R -i base_dir -i template_DE_stats_filename -i hallmark_DB_filename -i statistic -o template_enriched_pathways

source(paste(base_dir, 'generic_expression_patterns_modules/GSEA_analysis.R', sep='/'))
...
```
This ensures that `generic_expression_patterns_modules/GSEA_analysis.R` will be found in R. In general, it is not a good idea to use relative path when we are running the script in interactive mode (iPython included).

PR #30, opened 2020-10-02
Update readme with file locations

ajlee21:
This PR is adding the locations of the new data files to run just notebook 2, in case a user doesn't want to run notebook1 since it takes a long time to run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,11 @@ pip install -e .
 ```
 6. Navigate to either the `pseudomonas_analysis` or `human_analysis` directories and run the notebooks in order.
 
-*Note:* Running the `human_analysis/1_process_recount2_data.ipynb` notebook can take a while since the dataset is very large. If you would like to run only the analysis  (`2_identify_generic_genes_pathways.ipynb`) to generate the human analysis results found in the publication, you can update the config file to use the following file locations: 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,10 +54,11 @@ pip install -e .
 ```
 6. Navigate to either the `pseudomonas_analysis` or `human_analysis` directories and run the notebooks in order.
 
-*Note:* Running the `human_analysis/1_process_recount2_data.ipynb` notebook can take a while since the dataset is very large. If you would like to run only the analysis  (`2_identify_generic_genes_pathways.ipynb`) to generate the human analysis results found in the publication, you can update the config file to use the following file locations: 
+*Note:* Running the `human_analysis/1_process_recount2_data.ipynb` notebook can take a couple days to run since the dataset is very large. If you would like to run only the analysis notebook (`2_identify_generic_genes_pathways.ipynb`) to generate the human analysis results found in the publication, you can update the config file to use the following file locations:
```
dongbohu, line 57: `couple days to run` probably should be `couple of days` ("to run" is redundant because the sentence starts with "Running ...").

dongbohu (approved):
A minor issue. Please feel free to close it when it is addressed.

Issue #14, opened 2020-09-01
Manual mapping in "replace_ensembl_ids" function


dongbohu:
`replace_ensembl_ids()` function applies a `manual_mapping` dict:
https://github.com/greenelab/generic-expression-patterns/blob/23e67888e982bbbc3b0adf566604356e325a6bc5/generic_expression_patterns_modules/process.py#L57-L62
but it is applied before https://github.com/greenelab/generic-expression-patterns/blob/23e67888e982bbbc3b0adf566604356e325a6bc5/generic_expression_patterns_modules/process.py#L76-L78
so these four manual mappings don't seem to be included in the returned `gene_expression` dataframe, because the four new column names (`PLEKHG7`, `LINC00595`, `TAS2R45`, `CCL3L1`) are not indexed in `gene_id_mapping`.

dongbohu:
The bug can be fixed like this (`manual_mapping` is passed in as an argument):
```python
def replace_ensembl_ids(original_expression, gene_id_mapping, manual_mapping):
    updated_mapping = gene_id_mapping.loc[~gene_id_mapping.index.duplicated(keep="first")]

    # Apply manual mapping here:
    for ensembl_id, gene_symbol in manual_mapping.items():
        updated_mapping.loc[ensembl_id].hgnc_symbol = gene_symbol

    updated_expression = original_expression.iloc[:, ~original_expression.columns.str.contains("PAR_Y")] 
    updated_expression.columns = updated_expression.columns.map(updated_mapping["hgnc_symbol"])

    updated_expression = updated_expression.iloc[:, updated_expression.columns != ""]
    updated_expression = updated_expression.iloc[:, updated_expression.columns.notnull()]

    return updated_expression
```

ajlee21:
Thank you for finding this error. I will look into this. I'm not sure if switching the order as you suggested will necessarily fix things in this case. I think I might need to merge my `manual_mapping` and `gene_id_mapping` dictionaries and then apply them.

dongbohu:
@ajlee21: The following two lines in my proposed code actually merged `manual_mapping` with `gene_id_mapping` (I am using another variable `updated_mapping` to avoid modifying the input parameter `gene_id_mapping` by accident):
```python
for ensembl_id, gene_symbol in manual_mapping.items():
        updated_mapping.loc[ensembl_id].hgnc_symbol = gene_symbol
```

ajlee21:
[...]

PR #31, opened 2020-10-02
Simplify test cases

dongbohu:
This PR includes:
- a bug fix of a (deprecated) function in `proces.py`. More details on the bug is available at #14.
- updates on notebooks of the two test cases.
- new output file: `human_analysis/data/shared_genes_human.pickle`
- deleted `tests/data/input/shared_genes_human.pickle`, since it is not being used at all


generic_expression_patterns_modules/process.py
```diff
@@ -28,65 +28,68 @@ def replace_ensembl_ids(expression_df, gene_id_mapping):
...
-    # same hgnc symbol. Homologous sequences are paralogous
-    # if they were separated by a gene duplication event: if a gene in an
-    # organism is duplicated to occupy two different positions in the same
-    # genome, then the two copies are paralogous
 
-    # Remove paralogs
+    NOTE:
+    -----
+    This function is deprecated due to large memory usage: when `expression_df`
+    is a large dataframe, manuipulating it inside the moemory becomes very slow
```
ajlee21, line 35:
```suggestion
    is a large dataframe, manuipulating it inside the memory becomes very slow
```

ajlee21 (approved):
This looks great! Thank you so much for taking the time to update the github actions tests!

generic_expression_patterns_modules/process.py
dongbohu: Thanks

PR #34, opened 2020-10-07
Add analysis notebooks using Rani dataset

ajlee21:
In https://github.com/greenelab/generic-expression-patterns/pull/15 we found that our model trained using recount2 rna-seq data was able to recapitulate the same ranking of genes, based on DE analysis, found in the Crow et. al. paper using a manual curation process on an array compendium. However, our recount2 trained model was NOT able to get consistent ranking of Hallmark pathways, based on GSEA analysis, compared to Rani Powers et. al.

This PR is adding analysis notebooks to train a model using Rani Powers et. al. dataset. to determine if what is causing this inconsistency.
* Is there a bug in the implementation of the GSEA analysis?
* Does the context of the training dataset determine which pathways are found?


generic_expression_patterns_modules/process.py
```diff
@@ -850,11 +822,7 @@ def get_shared_rank_scaled(
 
 
 def compare_gene_ranking(
```
georgiadoing, line 824: when comparing gene rankings, are higher ranks (more expressed genes) given more weight in the correlation calculation? I'm not sure if it makes sense to do that, I just thought of it based on your comment above (line 801).

georgiadoing (commented):
I was less familiar with this notebook than the other I reviewed on the generic genes, so I don't think I followed quite as well, but I appreciated the results about generic pathways!

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,778 @@
...
+# ## Need to customize code from ponyo
+# 
+# The current simulation-related function in ponyo, `get_sample_ids` assumes that the user is using one of two different metadata files (one associated with the pseudomonas compendium and another associated with recount2). The compendium dataset we are using here has a slightly different format for their metadata file. 
+# 
+# Here we are temporarily writing our own function customized for this Powers et. al. dataset. But we will be updating ponyo to allow for different metadata files in the future. Issue in ponyo is [here](https://github.com/greenelab/ponyo/issues/18) 
+
+# In[5]:
+
+
+def get_sample_ids(experiment_id, dataset_name, sample_id_colname):
```
jjc2718, line 116: Maybe you could just have this function take the metadata file as an argument? Then it seems like this function would generalize to the cases you have in `ponyo` as well, but then the user has to provide the file path of the metadata, so it's a bit of a tradeoff. Up to you.

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,778 @@
...
+    return sample_ids
+
+
+# In[6]:
+
+
+import glob
+from keras.models import load_model
+
+def shift_template_experiment(
```
jjc2718, line 152: Is this the same as the function in `ponyo`, except with `metadata_filename` as an argument? Might be good to make a note of that, so you remember in the future.

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,778 @@
...
+    ref_gene_col,
+    ref_rank_col,
+    figure_filename
+)
+
+
+# **Takeaway:**
+# * Previously we compared gene ranks obtained from (recount2)-trained VAE model vs gene ranks obtained from manual curation using Crow et. al data. This [PR](https://github.com/ajlee21/generic-expression-patterns/blob/807377d76f63b6282c62255d7b160feb8585e0e2/human_analysis/2_identify_generic_genes_pathways.ipynb) shows that the correlation of gene ranks are very consistent.
+# 
+# * Here we are comparing gene ranks obtained from a (Powers et. al.)-trained VAE model vs gene ranks obtained from manual curation using Crow et. al. Based on this correlation plot there is a high correlation between those very high and low ranked genes -- high correlation at the extremes but there is a lot of noise in the middle.
```
jjc2718, line 564: So this is similar to what you found using the recount2 model, right? At least the pair plot looks similar to me (high correlation for very high and very low ranked genes, but noisy in the middle)

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,778 @@
...
+# * Next let's try to use the fraction of adjusted p-value < 0.05 for our method and re-compare (below)
+# 
+# **Takeaway:**
+# * Previously we compared pathway ranks obtained from (recount2)-trained VAE model vs pathway ranking based on manual curation using Powers et. al. This [PR](https://github.com/ajlee21/generic-expression-patterns/blob/807377d76f63b6282c62255d7b160feb8585e0e2/human_analysis/2_identify_generic_genes_pathways.ipynb) shows that there was no correlation.
+# 
+# * Here we are comparing pathway ranks obtained from a (Powers et. al.)-trained VAE model vs pathway ranking based on manual curation using Powers et. al datasets. Based on this correlation plot there is a high correlation.
+
+# **Conclusion:**
+# 
+# * We can get relatively similar generic genes training our VAE model on a cancer-specific dataset (Powers et. al.) and testing on a genenal dataset (Crow et. al.). These generic genes are not *that* context-specific at the extremes
```
jjc2718, line 774: Can you clarify what you mean by "similar" here? Do you mean the genes are similar between experiments, or between what you find using simulated data and what the paper found using real data?

I think you mean the latter (comparing the scatter plot above to the one in `human_analysis/2_identify_generic_genes_pathways.py`, it looks like your results here are much more correlated with the ones from Powers et al, which I'm guessing is what you're saying), but would be good to make this clear.

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,778 @@
...
+# 
+# * We get very different generic pathways training our VAE model on a general dataset (recount2) and testing on a cancer-specific dataset (Powers et. al.) -- see [this analysis](../human_analysis/2_identify_generic_genes_pathways.ipynb). But we get very similar generic pathways training our VAE model on a cancer-specific dataset (Powers et. al.) and testing on a cancer-specific dataset (Powers et. al.). This indicates that generic pathways are more context specific.
+# 
+# * I need to think about about why there is a difference in genes vs pathways.
```
jjc2718, line 778: Maybe the strong signal in pathways is driven by small expression changes in lots of downstream genes, so when you look on an individual gene level the results don't really reflect the change in context, but when you look at pathways as a whole you're better able to differentiate between contexts?

I think maybe this also makes sense if you think about it in the context of the paper Ben and I presented at journal club (https://www.biorxiv.org/content/10.1101/2020.07.30.228296v1): different experimental conditions lead to changes in the activity of signaling pathways, which lead to changes in the expression of the genes these signaling pathways regulate as a group. Since the genes are regulated as a group, you'd see (possibly small) coordinated changes in expression that aren't necessarily significant on the individual gene level.

Anyway, I'm just throwing some ideas out there (and I really like the idea of thinking about this on a pathway level, especially in the context of cancer - although I know you're firmly on team microbe and probably aren't too interested in cancer applications 😂 ). I think you'd need to do further analyses to really confirm (or deny) any of what I'm saying, but maybe something in here will help you form a mental model of how this is working, at least.

jjc2718 (approved):
Looks good! Just a few things, mostly conceptual comments/questions.

I'll approve because I think the code seems sound/correct, but feel free to wait for @danich1 to make comments as well.

human_cancer_analysis/nbconverted/1_process_data.py
```diff
@@ -0,0 +1,238 @@
+
+# coding: utf-8
+
+# # Process Powers et. al. data
```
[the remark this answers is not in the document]
ajlee21: Yes the notebook `1_process_data.ipynb` and `nbconverted/1_process_data.py` contain the same content. The `.py` file is just the notebook converted into a script so that you can easily comment in line for these types of PRs.

human_cancer_analysis/2_identify_generic_genes_pathways.ipynb
```diff
@@ -0,0 +1,2683 @@
...
+    "* Here we are comparing pathway ranks obtained from a (Powers et. al.)-trained VAE model vs pathway ranking based on manual curation using Powers et. al datasets. Based on this correlation plot there is a high correlation."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Conclusion:**\n",
+    "\n",
+    "* We can get relatively similar generic genes training our VAE model on a cancer-specific dataset (Powers et. al.) and testing on a genenal dataset (Crow et. al.). These generic genes are not *that* context-specific at the extremes\n",
```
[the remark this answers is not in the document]
ajlee21: So the steps are:
1. Train VAE on a compendium of gene expression data, which can include the template experiment (in our case the code is implemented in such a way that the template needs to be included in the training set).
2. Then the template experiment is used to simulate new experiments using that trained VAE.

I can add a comment to clarify this

generic_expression_patterns_modules/process.py
```diff
@@ -680,29 +681,27 @@ def process_raw_template(
...
-            sample_id = line.split('\t')[0]
+            sample_id = line.split("\t")[0]
             if idx == 0 or sample_id not in sample_ids_to_drop:
                 ofh.write(line)
 
 
 def normalize_compendium(
-        mapped_filename,
-        normalized_filename,
-        scaler_filename,
```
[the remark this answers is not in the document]
ajlee21: Yup! Something nifty that the scikit-learn library allows for these data transformations. I use the untransformation to perform the DE analysis

generic_expression_patterns_modules/process.py
ajlee21: Our gene rank = median abs(log2FC) across multiple simulated experiments. So high gene rank corresponds to those genes that were found to be DE across multiple simulated experiments. 

So the correlation is showing how our gene rank compares with the ranking of the same genes found in the Crow et. al. paper, which performed DE across 600 experiments and ranked genes by how frequently they were found to be DE.

I will try to update the comments to explain this better

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Yes i think its better to use the metadata file as input and what I plan to do when I update ponyo.

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Yes I hackily changed it for this one case since it would take a bit to update ponyo. I should update the name of the function to not confuse myself, good point

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Yes similar, but there was a bit less noise in the middle and more of a clearer correlation pattern. This correlation is comparing generic genes found using a cancer-specific training set vs generic genes found using a mixed/general training set. So it would seem that the highly and lowly generic genes are well correlated but not much in the middle when we have cross training sets

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: This is great insight! Thank you for thinking about this. I'm so glad that you find the results exciting! I definitely need to think more about this once the repo is re-organized. But I was also thinking about your paper after I got these results:) 

ps. I shall always be team microbe...especially given I started in viral genomics before grad school. But I'm honestly team everything:)

human_cancer_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Good point. I will clarify this

Issue #25, opened 2020-10-01
Output in "human_analysis" notebooks 


dongbohu:
The following two notebooks:

- `human_analysis/1_process_recount2_data.ipynb`
- `human_analysis/2_identify_generic_genes_pathways.ipynb`

don't have outputs right now. To make them more user-friendly, we may want to have some of the outputs included.

ajlee21:
Notebook were rerun to populate cell outputs in this PR: https://github.com/greenelab/generic-expression-patterns/pull/36

PR #36, opened 2020-10-10
Rerun recount2 notebooks

ajlee21:
This PR is rerunning the notebooks using all SRA data from recount2. No code was changed in this PR. This PR is re-running the notebooks to get the cell outputs that were previously cleared. This should address https://github.com/greenelab/generic-expression-patterns/issues/25

Minor changes include:
* Rename the analysis direction from `human_analysis` to `human_general_analysis` to make way for adding a new analysis where we're training only on cancer-specific gene expression data. As opposed to recount2, which contains general gene expression data.

Note:
* I tried to train the model on 20 epochs as opposed to 40, since 20 seems sufficient based on the [log plot](https://github.com/ajlee21/generic-expression-patterns/blob/recount2_rerun/human_general_analysis/logs/NN_2500_30/tybalt_2layer_30latent_hist.svg). When I was running the notebook the internet connection crashed so the model training progress was lost. I have made an issue [here](https://github.com/greenelab/generic-expression-patterns/issues/37). Since models were already trained and uploaded in this [PR](https://github.com/greenelab/generic-expression-patterns/pull/22), I used these files to run the analysis notebook.


dongbohu (approved):
Looks good!

Issue #29, opened 2020-10-01
Remove "get_wd()" and "set_wd()" when downloading recount2 data


dongbohu:
The following two functions in R:
https://github.com/greenelab/generic-expression-patterns/blob/47f763cc23723156173decc4dd3640ca25a683ca/generic_expression_patterns_modules/download_recount2_data.R#L12
and
https://github.com/greenelab/generic-expression-patterns/blob/47f763cc23723156173decc4dd3640ca25a683ca/generic_expression_patterns_modules/download_recount2_data.R#L140

use `get_wd()` and `set_wd(...)` because they call `download_study(...)` with default arguments, so the downloaded files will be saved in `./<project_id>/` directory.

According to https://rdrr.io/bioc/recount/man/download_study.html, we can use `outdir` argument to specify the location of downloaded files. So the function call will be like this:
```
project_dir <- paste(download_dir, project_id, sep='/')
download_study(project_id, outdir=project_dir)
...
```

This will not only make the code cleaner, but also avoid the possible danger of changing the working directory (see #24).

ajlee21:
This [PR](https://github.com/greenelab/generic-expression-patterns/pull/40) makes this update in the test notebooks as a test case. This change seems to work. Next we will update the other analysis notebooks in a future PR.

ajlee21:
Addressed in https://github.com/greenelab/generic-expression-patterns/pull/40

PR #40, opened 2020-10-19
Update github actions tests

ajlee21:
This PR updates the test notebooks to use the new modules added by Dongbo

Key changes can be found in `tests/1_process_recount2_data.ipynb` and `tests/2_identify_generic_genes_pathways.ipynb`


generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -14,10 +14,10 @@ get_recount2_template_experiment <- function(project_id,
                                              raw_template_filename) {
 
   # Save current working directory
-  original_wd = getwd()
+  #original_wd = getwd()
```
dongbohu, line 17: As I mentioned in #29, if you comment out these `getwd()` and `setwd()` lines, you have to call `download_study()` later like this:
```
project_dir <- paste(download_dir, project_id, sep='/')
download_study(project_id, outdir=project_dir)
```
otherwise the downloaded files would be saved in current working directory, whatever it is.

The first test case is not affected because it didn't process the downloaded template file. Instead, it uses the raw template file that already exists in the repo. 
Also note that changing this function will affect `human_general_analysis/1_process_recount2_data.ipynb`.

generic_expression_patterns_modules/download_recount2_data.R
```diff
@@ -60,7 +60,7 @@ get_recount2_sra_subset <- function(template_project_id,
   metadata <- all_metadata()
   write.table(
     metadata,
-    paste(base_dir, '/human_analysis/data/metadata/recount2_metadata.tsv', sep = ""),
+    paste(base_dir, '/human_general_analysis/data/metadata/recount2_metadata.tsv', sep = ""),
```
dongbohu, line 63: If you want, you can use:
```paste0(base_dir, '/human_general_analysis/data/metadata/recount2_metadata.tsv')```
to simply the code a little bit.

dongbohu (commented):
Two comments for you to consider.

generic_expression_patterns_modules/download_recount2_data.R
dongbohu: You can try to read the downloaded template file right after the downloading step and see whether it is located in `template_download_dir`. I bet it is not.

dongbohu:
By the way, I found it very hard to review the test cases. Is there a special reason why it has to be written in Jupyter notebook?

ajlee21:
> By the way, I found it very hard to review the test cases. Is there a special reason why it has to be written in Jupyter notebook?

Good question. Originally this was just a result of me being used to having the notebooks. Thinking about it more, I think one reason to use notebooks is to use the R magic. Unless you know of another way to handle the R calls..?

dongbohu:
@ajlee21: There are a few ways to call R script inside Python: 
https://stackoverflow.com/questions/19894365/running-r-script-from-python
`subprocess` is good when the R script accepts parameters; `rpy2` seems more flexible.

generic_expression_patterns_modules/download_recount2_data.R
ajlee21: That makes sense. Thank you for checking. I have updated the script

ajlee21:
Based on our discussion offline, I will leave these tests as jupyter notebooks since:
1. Its nice to get a cell breakdown of what cell failed to help debug
2. These notebooks are acting as "integration" tests that are testing the analysis pipeline is running correctly. In the future, if time, I will add unit tests to complement the existing tests

dongbohu (approved):
Looks good!

generic_expression_patterns_modules/download_recount2_data.R
ajlee21: Turns out that the default `sep` is a space. So I need to have sep either way so I kept `sep=/`

dongbohu: The function that I suggested was `paste0`, not `paste`. But it's no big deal.

PR #41, opened 2020-10-22
Add GSEA to Pseudomonas analysis notebook

ajlee21:
The main change in this PR is adding GSEA analysis to Pseudomonas dataset -- see `pseudomonas_analysis/2_identify_generic_genes_pathways.ipynb`. 

Majority of the code was copied from the GSEA cells in `human_general_analysis/2_identify_generic_genes_pathways.ipynb`.

The new features include:
- Function to parse kegg pathway data into gmt format
- Clean up `process.py` functions
- Update column names for summary tables to be more interpretable


generic_expression_patterns_modules/process.py
```diff
@@ -572,12 +574,50 @@ def process_raw_compendium_recount2(
...
+    # Read in pathway data
+    pa_pathway_DB = pd.read_csv(pathway_DB_filename, sep="\t", header=None)
+
+    # Drop extra column
+    pa_pathway_DB.drop(columns=[1], inplace=True)
+
+    # Make genes tab-separated
+    pa_pathway_DB[2] = pa_pathway_DB[2].str.split(";").str.join("\t")
+
+    # Need to temporarily write data to file in order
```
jjc2718, line 598: Not sure I quite follow why you need to write this file twice - is the extra `\` part of the tabs that you're adding? I don't see any forward slashes in the original file, but maybe I'm missing something.

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -304,6 +311,132 @@
...
+
+
+get_ipython().run_cell_magic('R', '-i base_dir -i template_DE_stats_filename -i adage_kegg_DB_processed_filename -i statistic -o template_enriched_pathways', "\nsource(paste0(base_dir, '/generic_expression_patterns_modules/GSEA_analysis.R'))\ntemplate_enriched_pathways <- find_enriched_pathways(template_DE_stats_filename, adage_kegg_DB_processed_filename, statistic)")
+
+
+# In[24]:
+
+
+print(template_enriched_pathways.shape)
+template_enriched_pathways[template_enriched_pathways['padj'] < 0.05].sort_values(by='padj')
```
jjc2718, line 347: Do you want to print the whole table here, or would the first few rows (`df.head()`) be enough?

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -304,6 +311,132 @@
...
+)
+
+
 # ### Pathway summary table
+
+# In[31]:
+
+
+# Create intermediate file: "<local_dir>/gene_summary_table_<col_to_rank_pathways>.tsv"
+summary_pathway_ranks = process.generate_summary_table(
```
jjc2718, line 421: Would it make sense to do a pair plot here of the rank in the template experiment and the average rank in the simulated experiments (a bit like the one you did for the human data, except I think you were comparing against the processed data from Powers et al)?

Or are you not interested in how similar the template and simulated ranks are for this experiment?

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
```diff
@@ -304,6 +311,132 @@
...
+# In[32]:
+
+
+# Create `pathway_summary_filename`
+summary_pathway_ranks.to_csv(pathway_summary_filename, sep='\t')
+
+
+# **Quick check:**
+# 
+# Looks like Ribosomes are found to be significantly differentiated in the template and across 25 simulated experiments. So this pathway looks to be generic, which is consistent with [Crow. et. al.](https://www.pnas.org/content/116/13/6491.abstract)
```
jjc2718, line 442: Crow et al. used human data, right? Did you expect to see some similarities in the generic pathways between humans and Pseudomonas?

Maybe ribosomes in particular make sense since the protein sequences are known to be highly conserved between species (so presumably the functions are too)? Just guessing here, but it could be interesting.

generic_expression_patterns_modules/process.py
ajlee21: I was trying to write a csv file where each row was a different length, because each pathway has a different number of genes. When pandas tries to write the df which has the list of genes as a string, it tries to "escape" the commas. I'm trying to purposely unescape them with this line.

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: For the pseudomonas dataset, we don't have a reference to compare against like with the human dataset. The difference between the template vs simulated ranks is summarized in the z-score, which says how much the template test statistic deviates from the simulated test statistic.

generic_expression_patterns_modules/process.py
```diff
@@ -572,12 +574,50 @@ def process_raw_compendium_recount2(
...
+    Note: Currently this function is specifically
+    customized to expect pathway_DB_filename = 
+    "https://raw.githubusercontent.com/greenelab/adage/master/Node_interpretation/pseudomonas_KEGG_terms.txt"
+
+    """
+    # Read in pathway data
+    pa_pathway_DB = pd.read_csv(pathway_DB_filename, sep="\t", header=None)
+
+    # Drop extra column
+    pa_pathway_DB.drop(columns=[1], inplace=True)
```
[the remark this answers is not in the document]
ajlee21: Really helpful! thanks

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
jjc2718: Ohhh, got it. Do you have any other ideas of ways to validate these results? Or has nobody studied generic gene expression in Pseudomonas (and you're trying to discover new things here)?

ajlee21: So the idea here was that we validated that our method can identify generic genes using the human datasets, and this pseudomonas analysis is more for discovery of specific and generic gene candidates that we want to test in the lab thanks to the Hogan lab

generic_expression_patterns_modules/process.py
jjc2718: Got it. But shouldn't the lists of genes be tab-separated strings (since you did a string join in line 596 above)?

I guess it's not a big deal if you're only doing this once for a relatively short file, but it feels to me like there's probably a way to do this without writing the file twice. Maybe just make a note to look into this further if it's slow in the future?

pseudomonas_analysis/nbconverted/2_identify_generic_genes_pathways.py
ajlee21: Yes, crow et. al. findings were based on human data.

This is a good point. I actually forgot that the data sources were different when I looked at the findings. So i think its not so much that the protein has the same function across species, so much as these proteins are found to be DE across a wide context in both humans and pseudomonas. This this [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2679180/) talks about all the extra roles they can play which is consistent with this finding.

generic_expression_patterns_modules/process.py
ajlee21: Sorry I mis-typed. That should have said "tabs". This was what I found: https://github.com/pandas-dev/pandas/issues/16298

So at least I don't think that is currently a way to do it

PR #43, opened 2020-10-30
Add multiplier analysis

ajlee21:
This PR is adding a new analysis notebook and supporting functions

This notebook examines why generic genes are "generic"? Specifically, this notebook is trying to answer the question: Are generic genes found to be associated with many multiplier LV (which were designed such that some of them are associated with known pathways and gene sets)

The main changes can be found in notebook: `multiplier_analysis/0_process_multiplier_model.ipynb` and `multiplier_analysis/1_get_multiplier_LV_coverage.ipynb`

Some minor updates that don't need review:
1. Add random seed to notebook
2. Update `paste` to `paste0`
3. Fixed config file metacolname = `run` instead of `gsm`
4. Re-run `human_general_analysis/` notebooks to get multiple simulation results to input into this multiplier analysis
5. Add new result data files


ajlee21:
Note: github actions tests are failing due to an update in the h5py package. This will be resolved with second release of ponyo package (https://github.com/greenelab/ponyo/pull/19)

multiplier_analysis/nbconverted/0_process_multiplier_model.py
```diff
@@ -0,0 +1,135 @@
...
+# In[9]:
+
+
+readRDS = ro.r['readRDS']
+
+
+# In[10]:
+
+
+multiplier_full_model = readRDS(os.path.join(multiplier_dir,
```
miltondp, line 97: Since you are using Python 3.7, in the future, I found the pathlib module great to handle paths in a cross-platform way (thanks again @danich1 !). You can replace `os.path.join` with the `Path` class, which provides very nice extra methods.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -0,0 +1,249 @@
...
+        dpi=300,
+    )
+
+
+# **Takeaway:**
+# * On average, specific genes cover fewer pathway-associated LVs compared to generic genes, which were found to be linked to all pathway-associated LVs.
+# * This difference in coverage is correlated with the fact that there 1-6 specific genes identified compared to the 6000 generic genes found.
+# * Some of the LVs that were only found to have generic genes (see [table](generic_only_LV_summary.tsv)) include mainly immune response pathways (monocytes, mast cell activation), wound healing (collagen formation), cell signaling (focal adhesion, integrin1) 
+# 
+# **Overall, it looks like generic genes are associated with many pathways, acting as *gene hubs*, which is why they are "generic"**
```
miltondp, line 249: Cool!

miltondp (approved):
Very nice, Alex! I left some comments, but I think all of them are minor, so I'm approving this PR.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+# ## Quick looks at the data
+
+# In[10]:
+
+
+# Get a rough sense for how many genes contribute to a given LV
+# (i.e. how many genes have a value > 0 for a given LV)
+(multiplier_model_z > 0).sum()
```
jjc2718, line 119: Maybe you could sort this, to get a sense of what the bounds are. Are there any LVs that are associated with only a few genes?

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+    
+    generic_cov_ls += list(generic_cov_i)
+    specific_cov_ls += list(specific_cov_i)
+    
+from matplotlib_venn import venn2
+venn2([set(generic_cov_ls), set(specific_cov_ls)],
+      set_labels=("Generic LVs", "Specific LVs")
+     )
+
```
jjc2718, line 169: So does the output here mean there are only 8 specific genes that are in the MultiPLIER results? And ~6000 generic genes? Are you at all surprised by these numbers (or am I interpreting them wrong)?

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -0,0 +1,249 @@
...
+        bbox_inches="tight",
+        transparent=True,
+        pad_inches=0,
+        dpi=300,
+    )
+
+
+# **Takeaway:**
+# * On average, specific genes cover fewer pathway-associated LVs compared to generic genes, which were found to be linked to all pathway-associated LVs.
+# * This difference in coverage is correlated with the fact that there 1-6 specific genes identified compared to the 6000 generic genes found.
```
jjc2718, line 246: This is a neat analysis! I'm a bit skeptical of drawing conclusions from only 1-6 specific genes, but the results for generic genes (of which there are many more) being associated with many pathways make intuitive sense.

Your box plot looked at how many LVs had at least one generic gene in them. Could you flip this around: for each gene, plot how many LVs it's associated with; i.e. how many LVs that gene is nonzero in? I'm curious what this distribution would look like.

jjc2718 (approved):
Good work! I just have a few small things to add to Milton's comments.

multiplier_analysis/nbconverted/0_process_multiplier_model.py
ajlee21: This is a great point! I started playing around with using this paths.py file after seeing other greenelab member's repos - its definitely a lot more organized for defining paths. The reason I didn't end up using it for this repo is that I didn't think it was intuitive to the user to change the python file to import if they wanted to switch configurations. Also, at the time, I didn't think you could define both file paths and variables in the same script which made this direction seem less clean.

Since this repo is getting wrapped up, I will likely not make this refactor. But something I will definitely try to use in my next repo

multiplier_analysis/nbconverted/0_process_multiplier_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+import rpy2.robjects as ro
+from rpy2.robjects import pandas2ri
+from rpy2.robjects.conversion import localconverter
+
+
+# In[2]:
+
+
+multiplier_dir = os.path.join("/home/alexandra/Documents/Data/Generic_expression_patterns/", "multiplier")
```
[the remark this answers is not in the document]
ajlee21: Good point. I saw this as an accessory one-off analysis and got lazy and manually added this path. These are good points and I'll make this change

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: Good suggestion. This was definitely a check to see if there were any LVs with 0 genes contributing, which there doesn't look to be, and to possibly filter those out

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: The number of specific and generic genes are reported per simulation run. There are 4 runs each generating 25 simulated experiments and calculating z-score statistics based on these 25 experiments. Over the 4 runs half the time 0 specific genes were found and the other times there was 1 or 6 specific genes found. I was surprised by the number of specific genes being so few but this might be a result of the threshold I am using for the z-score cutoff and something that can be adjusted in the future.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: Yes I agree with your skeptism completely. Its not surprising that there was more coverage for generic genes vs specific genes because there were many more generic genes found. What if we try to randomly sample a generic gene set that is of the same size as the number of specific genes?

generic_expression_patterns_modules/process.py
```diff
@@ -1593,3 +1594,140 @@ def plot_venn(degs_traditional, degs_specific, degs_generic):
...
+
+        # Get predicted specific DEGs using z-score cutoff
+        ls_specific_genes = list(
+            (data[(data["abs(Z score)"] > z_threshold)].set_index("Gene ID").index)
+        )
+        print(f"No. of specific DEGs using z-score: {len(ls_specific_genes)}")
+
+        # Get predicted generic DEGs using z-score cutoff
+        ls_generic_genes = list(
+            (data[(data["abs(Z score)"] < z_threshold)].set_index("Gene ID").index)
```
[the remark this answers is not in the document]
ajlee21: Good catch

generic_expression_patterns_modules/process.py
```diff
@@ -1593,3 +1594,136 @@ def plot_venn(degs_traditional, degs_specific, degs_generic):
...
+def process_generic_specific_gene_lists(ls_generic_genes, ls_specific_genes, LV_matrix):
+    """
+    This function returns the list of generic genes and specific genes
+    that were included in the multiplier analysis. 
+
+    This prevents indexing by a gene that doesn't exist and resulting in NA values
+    """
+    multiplier_genes = list(LV_matrix.index)
+
+    generic_genes_processed = list(set(multiplier_genes).intersection(ls_generic_genes))
```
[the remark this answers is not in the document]
ajlee21: Because of this linter I have this line break is automatically set this way.

generic_expression_patterns_modules/process.py
```diff
@@ -1593,3 +1594,136 @@ def plot_venn(degs_traditional, degs_specific, degs_generic):
...
+    # 0: shared
+    # 1: generic only
+    # 2: specific only
+
+    grp_dict = {0: "shared", 1: "generic_only", 2: "specific_only"}
+
+    for LV_grp in range(len(all_LVs)):
+        if len(all_LVs[LV_grp]) > 0:
+            # Parse LV id
+            LV_ids = [int(i.split("LV")[-1]) for i in all_LVs[LV_grp]]
```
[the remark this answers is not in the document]
ajlee21: Yes this format is consistent. I would not have thought to use replace here...interesting!

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: After matching the number of generic and specific genes, there is equal coverage of LVs. This also isn't surprising given that most LVs have at least 1000's genes that contribute to them, so the number of genes is making a difference. For the next analysis we will try to account for the amount of contribution of genes in addition to presence/absence

generic_expression_patterns_modules/process.py
```diff
@@ -1593,3 +1594,140 @@ def plot_venn(degs_traditional, degs_specific, degs_generic):
...
+    genes, based on the statistics contained within the
+    summary dataframes
+
+    Here genes are determined as generic or specific based on their
+    z-score (i.e. how much gene's differential expression differs between
+    template vs null set)
+    """
+    ls_genes = []
+    for file in list_files:
+        print(f"Reading data for {file}")
```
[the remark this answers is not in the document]
ajlee21: This is such a great idea! Something that i wish i knew about for an earlier analysis. I'll definitely keep this in mind moving forward. Here is just a short print statement so I don't think a logger is helpful here.

PR #45, opened 2020-11-11
Update multiplier analysis

ajlee21:
This PR is updating the recent multiplier analysis based on a couple of realizations:

1. Previously we only included LVs that were significantly associated with pathways, but the other LVs should be included since they may capture some biology but weren't found to be significant because the signals were rare or not specific enough.

2. Previously we defined generic and specific genes based on z-score, but this z-score is specific to the template experiment, so a high z-score indicated that the gene was specific to that template experiment. But we are interested in examining all specific vs all generic genes. So here we will use gene ranking instead.

For this analysis we will compare 1) the number of LVs generic genes are present vs all other genes are present, 2) the number of LVs that generic genes are highly weighted vs all other genes are highly weighted.

Overall it looks like generic genes contribute a little to many LVs versus other genes that contribute a lot to some LVs


ajlee21:
Note: I believe checks are failing because test data is updated in [PR#44](https://github.com/greenelab/generic-expression-patterns/pull/44). I think these checks should pass once that PR is merged in

ajlee21:
I will be presenting the results of this work at demo day, so hopefully that helps.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -7,7 +7,7 @@
 # 
 # The PLIER model performs a matrix factorization of gene expression data to get two matrices: loadings (Z) and latent matrix (B). The loadings (Z) are constrained to aligned with curated pathways and gene sets specified by prior knowledge [Figure 1B of Taroni et. al.](https://www.cell.com/cell-systems/pdfExtended/S2405-4712(19)30119-X). This ensure that some but not all latent variables capture known biology. The way PLIER does this is by applying a penalty such that the individual latent variables represent a few gene sets in order to make the latent variables more interpretable. Ideally there would be one latent variable associated with one gene set unambiguously.
```
jjc2718, line 8: Looks like this link is broken - the parentheses in the URL are causing issues. I think this will work:
```suggestion
# The PLIER model performs a matrix factorization of gene expression data to get two matrices: loadings (Z) and latent matrix (B). The loadings (Z) are constrained to aligned with curated pathways and gene sets specified by prior knowledge [Figure 1B of Taroni et. al.](https://www.cell.com/cell-systems/pdfExtended/S2405-4712\(19\)30119-X). This ensure that some but not all latent variables capture known biology. The way PLIER does this is by applying a penalty such that the individual latent variables represent a few gene sets in order to make the latent variables more interpretable. Ideally there would be one latent variable associated with one gene set unambiguously.
```
But you might want to test this. If it doesn't work this StackOverflow post might be helpful: https://stackoverflow.com/q/26698243

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -49,265 +51,249 @@
...
 
+# ### High weight LV coverage
 
 # In[14]:
 
 
-# Create table of unique generic LVs, unique specific LVs, shared LVs
-process.create_LV_df(generic_cov_ls, specific_cov_ls, multiplier_model_summary)
+# Quick look at the distribution of gene weights per LV
+sns.distplot(multiplier_model_z["LV2"])
```
jjc2718, line 173: You can probably add `kde=False` here, I don't think the KDE is really adding much

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -7,7 +7,7 @@
...
 # 
-# While the PLIER model was trained on specific datasets, MultiPLIER extended this approach to all of recount2, where the latent variables should correspond to specific pathways or gene sets of interest. Therefore, we will look at the coverage of generic genes versus specific genes across these MultiPLIER latent variables.
+# While the PLIER model was trained on specific datasets, MultiPLIER extended this approach to all of recount2, where the latent variables should correspond to specific pathways or gene sets of interest. Therefore, we will look at the coverage of generic genes versus other genes across these MultiPLIER latent variables, which represent biological patterns.
```
jjc2718, line 10: So just to be clear, what is the difference between "specific genes" and "other genes"? Are specific genes the ones that were found to be differentially expressed in one or a few experiments in your analyses, and "other" genes are just all the genes that weren't found to be generic (i.e. not necessarily DE in any experiment)? Might be good to add a brief description of this difference here in the header.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -49,265 +51,249 @@
...
-fig2.set_xlabel("Gene Type",fontsize=14)
-fig2.set_ylabel(textwrap.fill("Proportion of pathway-associated LVs covered", width=30),fontsize=14)
-fig2.tick_params(labelsize=14)
-fig2.set_title("Coverage of pathway-associated LVs", fontsize=16)
 
+# In[20]:
+
+
+# Plot coverage distribution given list of generic coverage, specific coverage
+highweight_fig = sns.boxplot(data=all_coverage_df,
```
jjc2718, line 244: So the interpretation of these plots would be that generic vs. other genes have nonzero loading values for roughly the same number of LVs (i.e. the distributions are the same), but the other genes tend to contribute highly to the LVs more often than the generic genes?

Do you think the difference in this plot would be significant if you did, say, a t-test?

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -49,265 +51,249 @@
...
+highweight_fig.set_ylabel(textwrap.fill("Number of LVs", width=30),fontsize=14)
+highweight_fig.tick_params(labelsize=14)
+highweight_fig.set_title("Number of LVs genes contribute highly to", fontsize=16)
+
+
+# ## Get LVs that generic genes are highly weighted in
+# 
+# Currently, I get high weight genes if the gene weight > quantile threshold for that specific LV. Each LV has a different threshold based on its own distribution. So each LV has the same number of high weight genes. 
+# 
+# Should I have normalized the data per LV before calculating which genes are high weight per LV?
```
jjc2718, line 261: Re. this:

> Should I have normalized the data per LV before calculating which genes are high weight per LV?

I think it would be worth trying. I guess what you would do is first normalize the columns of the MultiPLIER matrix (using, say, [StandardScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html)), then pick a cutoff and use it for all the columns.

My guess is that this will give you very similar results to what you have, though. If what you get is substantially different, that would mean some of the LV weight distributions are quite skewed (since using quantiles will "correct" for this to some degree, and normalization won't really). I don't necessarily think there's anything wrong with the way you're doing things now (or I think you can make a strong argument for it being correct, anyway), but trying the standardization approach as well might help you to put your results into context more.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
```diff
@@ -317,8 +303,9 @@
 
 
 # **Takeaway:**
-# * On average, specific genes cover fewer pathway-associated LVs compared to generic genes, which were found to be linked to all pathway-associated LVs.
-# * Warning: This difference in coverage is not surprising given the fact that there 1-6 specific genes identified compared to the ~1000 generic genes found.
-# * Some of the LVs that were only found to have generic genes (see [table](generic_only_LV_summary.tsv)) include mainly immune response pathways (monocytes, mast cell activation), wound healing (collagen formation), cell signaling (focal adhesion, integrin1) 
+# * Generic and other genes have are present in a similar number of LVs. This isn't surprising since the number of genes that contribute to each LV is <1000.
```
jjc2718, line 306:
```suggestion
# * Generic and other genes are present in a similar number of LVs. This isn't surprising since the number of genes that contribute to each LV is <1000.
```

jjc2718 (approved):
Looks good! Just a few small things (and some thoughts on your results).

generic_expression_patterns_modules/process.py
```diff
@@ -1610,141 +1612,135 @@ def get_gene_summary_files(data_dir):
...
-        set(multiplier_genes).intersection(ls_specific_genes)
-    )
+        dict_nonzero_coverage[gene_label] = LV_series
 
-    return generic_genes_processed, specific_genes_processed
+    return dict_nonzero_coverage
 
 
-def get_LV_coverage(ls_generic_genes, ls_specific_genes, LV_matrix):
+def get_highweight_LV_coverage(dict_genes, LV_matrix, quantile):
```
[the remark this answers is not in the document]
ajlee21: Good point. I'll add this

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: Yes that is correct. Other (or non-generic) genes is a superset that will include specific genes (i.e. genes that are DE in certain experiments) and housekeeping genes (i.e. genes that are consistently unchanged across experiments). I will definitely add text to clarify this. I can see my future self being confused.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: Yes your interpretation is correct. I was also debating about adding a significance statistic here.

multiplier_analysis/nbconverted/1_get_multiplier_LV_coverage.py
ajlee21: One thing Casey and I discussed was that it would be interesting to look at what LVs do these generic genes contribute a lot to. 

This question arose because I didn't think I could select LVs using the quantile approach since each LV had the same number of genes. At the time, I thought they all had the same number of high weight **generic genes**, but i realized today that they each have the same number of high weight genes **not necessarily generic** So I think I can just look at the proportion of high weight genes that are generic vs other to try to pull out LVs to look at - as in try to pull out LVs that have a higher proportion of generic high weight genes and look at the pathways associated with those LVs

generic_expression_patterns_modules/process.py
```diff
@@ -1610,141 +1612,135 @@ def get_gene_summary_files(data_dir):
     return glob(f"{data_dir}/generic_gene_summary_*")
 
 
-def process_multiplier_model_z(multiplier_model_z, multiplier_model_summary):
+def get_generic_specific_genes(summary_data, generic_threshold):
```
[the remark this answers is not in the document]
ajlee21: Thank you for the very thorough explanation. I was recently made aware of this option:) Since this threshold was customized to the correlation plot for this analysis, I will leave this function without a default for now.

PR #48, opened 2020-12-08
Make volcano plots for representative simulated experiments

ajlee21:
This PR is adding a notebook to create volcano plots to show the biological patterns that the simulation creates. In particular, showing that the structure of the volcano plots is consistent between template and simulated experiments and across simulated expreriments.

This PR also adds this new plot as a panel to the manuscript figures.

Note: The fonts are inconsistent for some reason I'm not sure of - when viewing the output figures on linux, some fonts are Roboto and others are Verdana. But when viewing the output figures on my windows, all the fonts are Verdana.

These volcano plots will need to be updated in a future PR based on meeting discussion. But these updated volcano plots will depend on changes made in https://github.com/greenelab/generic-expression-patterns/pull/50. So I want to merge these changes in the in term and I'll open a new PR in the future.

The main change to review in this PR can be found in [Viz_simulated_experiments.ipynb](https://github.com/greenelab/generic-expression-patterns/blob/41b445d747a5f21b767c9a7a0639003403a2a76a/human_general_analysis/Viz_simulated_experiments.ipynb). However this does not need a heavy review since it will be changing in the future.


human_general_analysis/nbconverted/Viz_simulated_experiments.py
```diff
@@ -0,0 +1,215 @@
...
+            "none": ".",
+            "DEG": "o",
+        },
+        palette=colors,
+        linewidth=0,
+        alpha=0.5,
+        )
+    
+    f.set_xlabel("log2 Fold Change", fontsize=14, fontname="Verdana")
+    f.set_ylabel("-log10(FDR adjusted p-value)", fontsize=14, fontname="Verdana")
```
danich1, line 105:
```suggestion
    f.set_ylabel(
      "-log10(FDR adjusted p-value)", 
      fontsize=14, 
      fontname="Verdana"
     )
```
Add white space for above and the next line below

danich1 (approved):
LGTM! Great use of svgs for high resolution figures! Just one change and good to merge.

PR #49, opened 2020-12-08
Add multiple enrichment methods 

ajlee21:
The goal of this PR was to add other multiple enrichment methods to demonstrate the our simulation method can be plugged into other pipelines easily.

The changes to this PR will depend on changes in this PR https://github.com/greenelab/generic-expression-patterns/pull/50. So for now I want to merge these changes in the in term and I'll open a new PR in the future to finish the work that this PR started.

The main change in this PR can be found in [Try_other_enrichment_methods.ipynb](https://github.com/greenelab/generic-expression-patterns/blob/8ef13425991e614e89333fda840216ee244fcab2/human_general_analysis/Try_other_enrichment_methods.ipynb). There is not much in this notebook. It mainly starts to outline and setup the changes to be made. This doesn't need a heavy review as this will be changed in the future.

### DIFF
--- a/human_general_analysis/nbconverted/Try_other_enrichment_methods.py
+++ b/human_general_analysis/nbconverted/Try_other_enrichment_methods.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Name
+# 
+# This notebook plugs in other gene set enrichment methods to demonstrate that our method, SOPHIE, can be inserted into different pipelines and work with other methods
+
+# In[1]:
+
+
+get_ipython().run_line_magic('load_ext', 'autoreload')
+get_ipython().run_line_magic('load_ext', 'rpy2.ipython')
+get_ipython().run_line_magic('autoreload', '2')
+
+import os
+import sys
+import pandas as pd
+import numpy as np
+import pickle
+
+from rpy2.robjects import pandas2ri
+pandas2ri.activate()
+
+from ponyo import utils
+from generic_expression_patterns_modules import calc, process
+
+np.random.seed(123)
+
+
+# In[2]:
+
+
+# Read in config variables
+base_dir = os.path.abspath(os.path.join(os.getcwd(), "../"))
+
+config_filename = os.path.abspath(
+    os.path.join(base_dir, "configs", "config_human_general.tsv")
+)
+
+params = utils.read_config(config_filename)
+
+
+# In[3]:
+
+
+# Load params
+local_dir = params["local_dir"]
+project_id = params['project_id']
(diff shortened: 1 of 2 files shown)


human_general_analysis/nbconverted/Try_other_enrichment_methods.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+import sys
+import pandas as pd
+import numpy as np
+import pickle
+
+from rpy2.robjects import pandas2ri
+pandas2ri.activate()
+
+from ponyo import utils
```
danich1, line 24: What tensorflow version are you using? If it is less than 2.x.x, your ponyo code might become outdated. ([Example Here](https://github.com/KrishnaswamyLab/SAUCIE/pull/38) Really tiny issue in the overall scheme, but I wanted to just let you know.

human_general_analysis/nbconverted/Try_other_enrichment_methods.py
```diff
@@ -0,0 +1,92 @@
...
+from generic_expression_patterns_modules import calc, process
+
+np.random.seed(123)
+
+
+# In[2]:
+
+
+# Read in config variables
+base_dir = os.path.abspath(os.path.join(os.getcwd(), "../"))
```
danich1, line 34: Might want to try and use pathlib. I know broken record, but it definitely makes os path generation a lot easier to deal with.

danich1 (approved):
LGTM! Left tiny comments for over grand-scheme, but no changes necessary on this PR.

human_general_analysis/nbconverted/Try_other_enrichment_methods.py
ajlee21: Thanks for looking out. I remember having some conflicts with other dependencies that arose when i previously tried to upgrade tensorflow. But it has been a while since I looked at this so I'll make an issue in my ponyo repo, which is using v<2.x.x for tensorflow

human_general_analysis/nbconverted/Try_other_enrichment_methods.py
ajlee21: Not a broken record. I definitely agree with using paths instead of the way I currently have my config. The legacy reason for having the config this way is 1) to be consistent with how ponyo is, which this repo uses and 2) it seemed more user friendly at the time to have a tab-delimited file and have that passed into the notebook as opposed to "importing a python file. I also wasn't aware, at the time when I created the config, that I could define both variables and paths in the same file. But rest assured that my other [newer repo](https://github.com/greenelab/core-accessory-interactome) uses paths:)

PR #50, opened 2020-12-10
Run new experiment through pipeline

ajlee21:
This PR is pretty narly looking so I'll try to explain what happened:)

Initially this PR was adding a notebook that allows users to run an experiment of interest through the pipeline to identify specific genes of interest. This notebook is using existing human VAE model to embed this new experiment and simulate experiments.

The changes associated with this can be found here (**no review required** since I will submit a new PR to rerun this analysis and clean up the code after the addressing the issues below):
1. `new_experiment/find_generic_genes_in_new_experiment.ipynb`: notebook that allows users to run their own experiment through SOPHIE
2. Updated ReadMe to describe how to run your own analysis.
3. Scripts in `new_experiment_process.py`: supporting functions for this new analysis

After presenting the initial results this notebook generated, we noticed an issue that is described in `new_experiment/debug.ipynb`. 

The main changes to review are making corrections to the code to address the issue found as well as additional changes requested during the meeting:
1. `new_experiment/debug.ipynb`: notebook that explains and illustrates the issue.
2. `process.py`, `stats.py`, 'ranking.py`: Update and re-organize supporting scripts to 1) consolidate multiple functions and process data more neatly; 2) add functionality to filter lowly expressed genes when using RNA-seq data; 3) Correct summary tables to use raw z-score since we want specific genes to be those that change MORE than in the null set.
3. Update test notebooks based on updates to scripts.

The following additional changes were made but **do not require a review**:
1. Re-ran notebooks in `human_general_analysis`. The scaler transform, normalized compendium file and VAE models were consistent with each other and generated by Dongbo. However the mapped compendium file was not because this mapped file was generated later by me. The training notebook crashed mid-run so I used the previous scaler transform, normalized compendium and VAE models. This inconsistency is relevant to the new notebook introduced in this PR because it needs to use the mapped compendium file to match the feature space and use the corresponding VAE models to embed the data.
2. After making the corrections to the code, the analysis notebooks in `human_general_analysis` and `human_cancer_analysis` were re-run since the new code will update the validation.


generic_expression_patterns_modules/stats.py
```diff
@@ -0,0 +1,239 @@
...
+    if metadata_sample_ids.equals(expression_sample_ids):
+        print("sample ids are ordered correctly")
+    else:
+        # Convert gene expression ordering to be the same as
+        # metadata sample ordering
+        print("sample ids don't match, going to re-order gene expression samples")
+        expression = expression.reindex(metadata_sample_ids)
+
+    # Save
+    if out_expression_filename != None:
```
jjc2718, line 145: When I'm comparing against None, I typically try to use `is not None` rather than `!= None`, although in most cases it doesn't matter. (Also, `is not None` is slightly faster)

See here for some cases when it does matter and an explanation: http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html .

new_experiment/nbconverted/debug.py
```diff
@@ -0,0 +1,405 @@
...
+# In[41]:
+
+
+template_nonrecount2_nofilter_DE_stats.loc["SSTR5.AS1"]
+
+
+# **Takeaways:**
+# * Case 1: genes with only simulated statistics is because template experiment has all 0 counts
+# * Case 2: genes with fewer than 25 simulated experiments, some simulated experiments have all 0 counts for those genes
+# * Case 3 (only found using non-recount2 template): genes with missing p-value in template experiment have NaN output from DESeq2, which indicates those genes that have been automatically filtered by DESeq2 as very likely to not have no significance: https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#pvaluesNA
```
jjc2718, line 395: Interesting! So it turned out that none of these cases were due to issues with name resolution/name mapping, just due to how DESeq is handling things or cases with all 0 gene counts?

new_experiment/nbconverted/debug.py
```diff
@@ -0,0 +1,405 @@
...
+# 
+# 
+# **Proposed solution:**
+# * Case 1 and 2: Remove genes with 0 counts across all samples. I will also add an option to additional remove genes that have mean counts < user specified threshold (for Jim's case). This should get rid of those rows with missing values in template statistic columns. These genes will be missing in the validation against Crow et. al. so I'll need to ignore these genes in the rank comparison. Any gene removed from the simulated experiments will get a lower `number of simulated experiments` reported so I will need to document this for the user.
+# * Case 3: (option 1) I can change the parameter setting to turn off this autofiltering and it will perform all tests and report them. (option 2) Replace padj values = "NA" with "Filtered by DESeq2" and document that this means that DESeq2 pre-filtered these genes as likely not being signicant to help increase detection power. For now I am using option 1. DESeq2 documentation states:
```
jjc2718, line 400: Will Option 1 take a lot longer to run?

new_experiment/nbconverted/debug.py
```diff
@@ -0,0 +1,405 @@
...
+# 
+# 
+# **Proposed solution:**
+# * Case 1 and 2: Remove genes with 0 counts across all samples. I will also add an option to additional remove genes that have mean counts < user specified threshold (for Jim's case). This should get rid of those rows with missing values in template statistic columns. These genes will be missing in the validation against Crow et. al. so I'll need to ignore these genes in the rank comparison. Any gene removed from the simulated experiments will get a lower `number of simulated experiments` reported so I will need to document this for the user.
+# * Case 3: (option 1) I can change the parameter setting to turn off this autofiltering and it will perform all tests and report them. (option 2) Replace padj values = "NA" with "Filtered by DESeq2" and document that this means that DESeq2 pre-filtered these genes as likely not being signicant to help increase detection power. For now I am using option 1. DESeq2 documentation states: 
+# ```
```
jjc2718, line 401: I think the formatting for this block is a little off - not sure why, though. Maybe you need a double line break to break out of the bullet point?

jjc2718 (approved):
Nice work! This definitely seems like quite a mystery - good job tracking down the issues.

I focused my review on the files you listed, although I had a few small comments on other things (mainly the README, just a few suggestions to improve clarity).

generic_expression_patterns_modules/stats.py
```diff
@@ -0,0 +1,239 @@
...
+        be output to the same input filename   
+
+    """
+
+    # Read data
+    expression = pd.read_csv(expression_filename, sep="\t", index_col=0, header=0)
+    process_metadata = pd.read_csv(
+        process_metadata_filename, sep="\t", index_col=0, header=0
+    )
+    grp_metadata = pd.read_csv(grp_metadata_filename, sep="\t", header=0, index_col=0)
```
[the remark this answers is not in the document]
ajlee21: I agree with your change. This is something that seems to get auto-corrected by the linter though

generic_expression_patterns_modules/ranking.py
```diff
@@ -0,0 +1,818 @@
...
+    This function takes the absolute value of columns=[`logFC`, `t`, `NES`].
+    For ranking genes or pathways, we only care about the magnitude of the change for
+    the logFC, t, NES statistic, but not the direction.
+
+    The ranking for each gene will be based on the mean absolute value of either
+    logFC, t, NES statistic, depending on the user selection
+    """
+    if "logFC" in simulated_DE_stats_all.columns:
+        simulated_DE_stats_all["logFC"] = simulated_DE_stats_all["logFC"].abs()
+    elif "log2FoldChange" in simulated_DE_stats_all.columns:
```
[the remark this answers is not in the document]
ajlee21: There should be either "logFC" or "log2FoldChange" depending on the DE method that is used (limma vs DESeq)

new_experiment/nbconverted/debug.py
ajlee21: After rerunning this it doesn't seem like its adding a significant amount of time. Though I haven't performed any rigorous testing given the time constraint to complete this analysis

new_experiment/nbconverted/debug.py
ajlee21: Yes. I assumed that there would be some issue with the name mapping given that this issue only appeared in the analysis using recount2 data. And the pipelines where similar expect this mapping step so that was my initial guess. But yes it does look like the issues are with DESeq. Which I think is an easier problem to solve compared to having to track down any mapping issues that can be a bit of a rabbit hole.

new_experiment/nbconverted/debug.py
ajlee21: Not sure what happened with this cell either. Just tried to do something else to fix the text though

generic_expression_patterns_modules/stats.py
```diff
@@ -0,0 +1,239 @@
...
+    metadata_sample_ids = grp_metadata.index
+    expression_sample_ids = expression.index
+
+    if metadata_sample_ids.equals(expression_sample_ids):
+        print("sample ids are ordered correctly")
+    else:
+        # Convert gene expression ordering to be the same as
+        # metadata sample ordering
+        print("sample ids don't match, going to re-order gene expression samples")
+        expression = expression.reindex(metadata_sample_ids)
```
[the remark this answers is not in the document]
ajlee21: Very good point! Always good to try to add more check especially given how big this repo is getting

generic_expression_patterns_modules/stats.py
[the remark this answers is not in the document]
jjc2718: Ok, don't worry about it if the linter thinks I'm wrong. I wonder why it doesn't like it!

generic_expression_patterns_modules/ranking.py
```diff
@@ -0,0 +1,818 @@
...
+        elif data_type.lower() == "gsea":
+            simulated_stats_file = os.path.join(
+                local_dir,
+                "GSEA_stats",
+                "GSEA_stats_simulated_data_" + project_id + "_" + str(i) + ".txt",
+            )
+
+        # Read results
+        simulated_stats = pd.read_csv(
+            simulated_stats_file, header=0, sep="\t", index_col=0
```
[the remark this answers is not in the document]
ajlee21: Good catch! Looks like it works

generic_expression_patterns_modules/multiplier.py
```diff
@@ -0,0 +1,308 @@
...
+import pandas as pd
+
+
+# Functions for multiplier analysis
+def get_gene_summary_file(data_dir):
+    """
+    This function returns a list of file paths for all files
+    of the form `generic_gene_summary_*` in data_dir.
+
+    Note: There should only be one file that has this form.
```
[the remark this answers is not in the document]
ajlee21: It looks like this function is actually unnecessary now. I was originally planning to run this across all the summary results. But I don't think that's necessary now as I'm sure this trend will be present using different template experiments and the user can specify which config to use in that case. So I will remove this.

ajlee21:
Note: pytests are currently failing because of an error with recount2 download. I have tried a different experiment id and this did not resolve things. Will wait to see if this issue gets resolved before updating the pytests

PR #52, opened 2021-01-06
Add new bladder cancer result files

ajlee21:
No code changes in this PR

This PR is re-running the analysis to identify specific genes that are associated with resistance to bladder cancer treatments (cis-resistant vs parental, gem-resistant vs parental or cis-gem-resistant vs parental). Previously we pre-filtered genes that had low counts. However this resulted in many specific genes (i.e. those with high z-score) with a null set composed of only a few simulated experiments. Here we are removing this threshold and re-running the analysis.

No intensive code review necessary as this PR is mainly adding new data files.


jjc2718 (approved):
Looks good to me!

Does this change the interpretation of any of your results at all (e.g. from your presentation [here](https://docs.google.com/presentation/d/1WfbsharC50gK0Hccvzngvfr5Aje-zipuDkv4OI5e4ZA/edit#slide=id.gb1616fffac_0_1))? I'm guessing you've worked through these results already with Casey and your collaborators, I'm mostly just curious for my own benefit.

ajlee21:
Good question. The results don't change if we don't apply the filter. Overall it looks like using an unseen experiment as our template compared to one that exists in our training compendium, our simulation has a little bit more difficulty capturing this pattern. So the volcano plots are not as consistent in this case, but we believe that our simulation approach should still yield roughly similar simulated experiments that we can use to identify candidate specific genes. And hopefully in the future this method can be tuned when we are able to learn more about the latent space structure and what its capturing.

PR #53, opened 2021-01-07
Compare generic genes across two template experiments

ajlee21:
This PR contains the following changes:

1. Updates the volcano plot in the figure manuscript to use a different recount2 experiment as a template, one with a less strong signal that is more representative of the type of data that users would be working with. (See `figure_generation/output/figure_1.svg` code to generate the new volcano plot involved re-running `human_general_analysis/2_identify_generic_genes_pathways.ipynb` using different project id)

2. Adds a new analysis notebook (`compare_experiments/compare_generic_genes.ipynb`) that will compare the generic genes identified using one template experiments vs another. We found that there are some genes that are generally generic (i.e. those that are concordant between these two experiments) and there are also genes that are generic within the given context of the specific experiment. Having different gene expression distributions doesn't seem to account for the differences in gene ranking, possibly suggesting that level of gene expression activity doesn't matter as much as the overall patterns perhaps.


compare_experiments/nbconverted/compare_generic_genes.py
```diff
@@ -0,0 +1,210 @@
...
+# Distribution of concordant genes in template experiment 1
+template1_mean = template_1.mean()
+
+print("Percent concordant genes with 0 expression in template 1:", 
+      len(template1_mean[concordant_genes].loc[template1_mean[concordant_genes]==0])/len(template1_mean[concordant_genes]))
+
+print("Percent nonzero concordant genes in template 1:", 
+      len(template1_mean[concordant_genes].loc[(template1_mean[concordant_genes]>0) &
+                                           (template1_mean[concordant_genes]<1000)])/len(template1_mean[concordant_genes]))
+sns.distplot(template_1.mean()[concordant_genes], kde=False)
```
jjc2718, line 152: Could you add axis labels here? It's not quite clear what you're plotting at first glance (although I figured it out by reading the code/comments).

Same comment for the next few plots.

jjc2718, line 152: Also, looks like it might be better to plot log(count) on the y-axis here, instead of the raw count (which seems pretty skewed toward 0).

compare_experiments/nbconverted/compare_generic_genes.py
```diff
@@ -0,0 +1,210 @@
...
+sns.distplot(template_2.mean()[discordant_genes], kde=False)
+
+
+# **Takeaway:**
+# 
+# Doesn't appear to be much of a difference between the distribution of average gene expression values for these two experiments. 
+# 
+# Theoretically, I would expect the scenario where a gene is lowly expressed in the context of template experiment 1 and therefore not found to be generic. But this same gene could be found to be generic in the context of template experiment 2 if it is more expressed. Its possible that differences in gene expression distribution can change which genes are found to be generic given that the simulation is producing experiments with a similar context. 
+# 
+# In this case, despite having similar gene expression distributions there are still many differences in gene ranking. This suggests to me that level of gene expression activity doesn't matter as much as the overall patterns perhaps.
```
jjc2718, line 210: This makes sense looking at the distributions, doesn't it? If almost all genes have expression levels very close to 0, then small fluctuations in the expression of some genes could lead to large changes in rank without changing the overall (skewed) distribution very much.

It's definitely interesting to know that the overall distributions are the same (which is what you're showing here), but I think looking at a log plot (or even a log-log plot, since the x-axis has a pretty large range too) could help to elucidate some of the differences between lowly expressed genes that seem to be driving the changes in rank.

jjc2718 (approved):
Looks good! Seems like you're making good progress - excited to read the paper once it's done.

A few comments on `figure_generation.ipynb` (diff is too large to comment inline):

* `svgutils` is super cool, I didn't know you could make these kinds of publication-ready figures programmatically! I always just do it in Inkscape, but this is way more reproducible.

* Do you want to update the Python script in `nbconverted` for this file? It doesn't look like that changed at all.

* For the second notebook cell, you should be able to replace this:
```python
if not os.path.exists(output_directory):
    os.makedirs(output_directory)
```
with this:
```python
os.makedirs(output_directory, exist_ok=True)
```

* Just a suggestion (feel free to take it or leave it), but in your `make_figure_panel` function it might be clearer to use the `x_` and `y_` parameters keyword arguments (i.e. call them by keyword instead of positionally). I know for me personally, it would be hard to remember the order of them. So you could instead call the function like this:
```python
panel_1a = make_figure_panel(
    "fig1A.svg",
    scale_x_input=2.5, 
    scale_y_input=2.5,
    x_loc=30,
    y_loc=10
)
```
and you don't have to change anything about the function definition (although you can add defaults if you want). Again, just something that would simplify things for me, but if you're fine with remembering the order don't worry about it.

ajlee21:
Thank you for the tips on the figure_generation notebook. I think these will definitely make things more clear! I'm so happy I learned about the svgutils..courtesy of @danich1

compare_experiments/nbconverted/compare_generic_genes.py
ajlee21: This is a great point. Thank you for thinking about this. After scaling by logs, we observe a slight shift showing that concordant genes are more lowly expressed compared to discordant genes, but most genes are still predominantly lowly gene expression. This would follow what you're saying -- If most genes have expression levels very close to 0, then small fluctuations in the expression of some genes could lead to large changes in rank without changing the overall expression distribution.

I'll need to continue to think about this some more. But It would make sense to me that the patterns as opposed to the gene expression levels matter more for differentiating gene ranking. And that your logic for why similar distributions of expression can lead to differences in rank can explain why this is the case.

PR #54, opened 2021-01-08
Add enrichment analyses

ajlee21:
The goal of this PR was to add other multiple enrichment methods to demonstrate the our simulation method can be plugged into other pipelines easily.

These changes were started in a previous [PR](https://github.com/greenelab/generic-expression-patterns/pull/49), but was prematurely closed due to other changes that this PR depended on.

This PR has the following changes:
1. Add new script `other_enrichment_methods.R` to run other enrichment analysis methods
2. Add directory with notebooks to simulate data and run different enrichment analyses `other_enrichment_methods/`
3. Updated previous scripts, analysis and test notebooks to allow for an additional parameter for user to specify enrichment method.


figure_generation/nbconverted/figure_generation.py
```diff
@@ -20,8 +20,7 @@
 
 # Directory of output figures
 output_directory = "output/"
-if not os.path.exists(output_directory):
-    os.makedirs(output_directory)
+os.makedirs(output_directory, exist_ok=True)
```
jjc2718, line 23: It looks like some of the changes in this notebook were made in your last PR (or at least I remember reviewing them before). Maybe merging in your `master` branch would help?

generic_expression_patterns_modules/other_enrichment_methods.R
```diff
@@ -0,0 +1,161 @@
...
+# Load libraries
+library("fgsea")
+library("limma")
+library("GSVA")
+library("DEFormats")
+library("edgeR")
+library("DESeq2")
+library("clusterProfiler")
+
+find_enriched_pathways_ROAST <- function(expression_filename,
```
jjc2718, line 15: Maybe it would be helpful to comment inside of each of these functions with a citation to the paper where each method is proposed or described (or refer people to the notebook `2_apply_enrichment_method.ipynb` where you have a nice description of each of them)?

generic_expression_patterns_modules/ranking.py
```diff
@@ -35,7 +37,9 @@ def concat_simulated_data(local_dir, num_runs, project_id, data_type):
         Number of simulated experiments
     project_id: str
         Project id to use to retrieve simulated experiments
-    data_type: 'DE' or 'GSEA'
+    data_type: 'DE' or 'GSA'
```
jjc2718, line 40: Does 'GSA' here just mean some kind of "gene set analysis" (i.e. enrichment analysis)?

generic_expression_patterns_modules/ranking.py
```diff
@@ -817,3 +862,148 @@ def process_and_rank_genes_pathways(
...
+
+
+def format_enrichment_output(
+    local_dir, project_id, enrichment_method, pathway_names, num_runs
+):
+    """
+    This function formats the output from the different enrichment methods
+    to rank and summarize pathway results
+
+    1. GSVA, which returns a matrix that is
```
jjc2718, line 874: This is helpful! Good to have a "key" for what each of the enrichment methods returns, and what you need to change to make them all look (roughly) the same for your analyses.

generic_expression_patterns_modules/ranking.py
```diff
@@ -817,3 +862,148 @@ def process_and_rank_genes_pathways(
...
+    3. ORA returns a matrix that is gene set x statistics.
+    We need to re-label index column as "pathway"
+
+    Arguments
+    ----------
+    local_dir: str
+        path to local machine where output file will be stored
+    project_id: str
+        Experiment identifier
+    enrichment_method: "GSVA", "ROAST", "CAMARA", "ORA"
```
jjc2718, line 888:
```suggestion
    enrichment_method: "GSVA", "ROAST", "CAMERA", "ORA"
```

human_general_analysis/nbconverted/Viz_simulated_experiments.py
```diff
@@ -155,29 +155,53 @@ def make_volcano_plot_simulated(
...
-            markers={
-                "none": ".",
-                "DEG": "o",
-            },
-            palette=colors,
-            linewidth=0,
-            alpha=0.5,
-            legend=False,
-            ax=axes[i],
+        if i == 0:
```
jjc2718, line 158: I think these changes were also from the previous PR (merging in master should fix).

other_enrichment_methods/nbconverted/1_simulate_data.py
```diff
@@ -0,0 +1,219 @@
...
+    stats.process_samples_for_DESeq(
+        simulated_filename,
+        metadata_filename,
+        out_simulated_filename,
+        count_threshold,
+        sample_id_metadata_filename,
+)
+
+
+# ### DELETE??? Differential expression analysis
```
jjc2718, line 177: Do you want to delete this? It looks like you're using the results in `2_apply_enrichment_method.ipynb` (unless I'm mistaken).

other_enrichment_methods/nbconverted/2_apply_enrichment_method.py
```diff
@@ -0,0 +1,231 @@
...
+
+# Output file
+pathway_summary_filename = os.path.join(
+    base_dir, 
+    dataset_name, 
+    f"generic_pathway_summary_{project_id}_{enrichment_method}.tsv"
+)
+
+
+# ## Enrichment methods
```
jjc2718, line 100: This is a great description!

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
```diff
@@ -0,0 +1,223 @@
...
+    "ORA": "p.adjust"
+}
+
+
+# ## Get pathway summary data
+
+# In[5]:
+
+
+# Pathway summary files
```
jjc2718, line 64: So this requires you to have run the previous notebook for all of the different GSA methods, right? That's fine, but it might be good to make a note of that somewhere (since people might run into errors if they only want to run it for one method, etc)

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
```diff
@@ -0,0 +1,223 @@
...
+# In[15]:
+
+
+plot_significance_vs_ranking(roast_pathway_summary, "ROAST", "FDR (BH)", "ROAST_pathway_ranking.svg")
+
+
+# In[16]:
+
+
+plot_significance_vs_ranking(camera_pathway_summary, "CAMERA", "FDR (BH)", "CAMERA_pathway_ranking.svg")
```
jjc2718, line 211: These look like they're in alphabetical order (i.e. "adipogenesis", "allograft rejection", "androgen response", etc). Is that just because there are a lot of significant pathways/gene sets (so it's just displaying the first few in alphabetical order), or does the order that gene sets are tested in matter for CAMERA?

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
```diff
@@ -0,0 +1,223 @@
...
+        transparent=True,
+        pad_inches=0,
+        dpi=300,
+    )
+
+
+# In[13]:
+
+
+plot_significance_vs_ranking(gsea_pathway_summary, "GSEA", "adjusted p-value (BH)", "GSEA_pathway_ranking.svg")
```
jjc2718, line 193: So is the "ranking" here on the y-axis the average DE rank of the gene set over simulated experiments? So you'd expect them to be somewhat correlated to the GSA enrichment statistics, but not perfectly correlated in all cases?

I know you've used these rankings before, just trying to remind myself of what they mean.

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
```diff
@@ -0,0 +1,223 @@
...
+# In[17]:
+
+
+plot_significance_vs_ranking(ora_pathway_summary, "ORA", "adjusted p-value (BH)", "ORA_pathway_ranking.svg")
+
+
+# **Takeaway:**
+# * Here are the results demonstrating that different enrichment methods can easily be plugged into our simulation workflow to identify generic gene sets
+# * Overall, it appears that the most enrichment gene sets are also those that are most commonly found to be enriched. However, depending on the enrichment method, generic gene sets will vary slightly due to the different assumptions and modeling procedures. More details about the methods can be found in the [previous notebook](2_apply_enrichment_method.ipynb)
+# * ROAST performs a self-contained test, which assesses the relevance of an individual biological process to the experiment at hand without reference to other genes in the genome. This lack of reference might account for why the scores are uniform across the different gene sets.
```
jjc2718, line 223: Yeah, it's weird that ROAST doesn't distinguish between gene sets at all! Wouldn't the rankings (y-axis) be different between gene sets, though? Or are the gene sets you're using different for this experiment?

jjc2718 (approved):
Looks good! My main point of confusion is just trying to figure out what the plots mean in `3_viz_pathway_ranking.ipynb`, but otherwise everything seems fine to me.

figure_generation/nbconverted/figure_generation.py
ajlee21: So these changes are because I merged in my master branch with this one to resolve some conflicts. So those are indeed changes from previously merged in PRs.

generic_expression_patterns_modules/ranking.py
ajlee21: Yea. So "GSA" refers to gene set analysis where as "GSEA" is a particular method. At the time GSEA was the only enrichment method I was using so that was why I had it hard coded. But now that there is the possibility of using other methods I wanted to update the name

human_general_analysis/nbconverted/Viz_simulated_experiments.py
ajlee21: See previous comment

other_enrichment_methods/nbconverted/1_simulate_data.py
ajlee21: good catch.

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
ajlee21: This is the best explanation I could come up with about the ROAST results. But all the methods used the same gene sets. My guess is that this method isn't appropriate here for some reason...maybe the number of gene sets is too many. I'll have to think more about this but I believe I've implemented it correctly.

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
ajlee21: Yes that correct. So higher ranking = more generic = pathways that are enriched across many of the simulated experiments.

The plots are showing the enrichment score in the template/real experiment of interest vs genericness of pathways.

It appears that the most enrichment gene sets are also those that are most commonly found to be enriched/generic.

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
ajlee21: Looks like by default the table is ordered by the `Rank (Real)` column, which is the rank  of the pathway (based on the enrichment) in the template experiment. I don't think the order should matter as each of the gene set tests are performed independently from what i can tell

other_enrichment_methods/nbconverted/3_viz_pathway_ranking.py
jjc2718: Yeah - especially in CAMERA it seems to be a very linear relationship. Interesting!

PR #55, opened 2021-01-08
Clean up and redo Pseudomonas analysis

ajlee21:
This PR is cleaning up and redoing the pseudomonas analysis using the new modules, where we have fixed the z-score issue.

No heavy review is required for this. The following things were changed:
1. Updated pseudomonas analysis notebooks to use new modules (`2_identify_generic_genes_pathways.ipynb` run with different config and metadata settings)
2. Archived files
3. Add new result files (`generic_gene_*` and `generic_pathway_*`)
4. Look into why volcano plots are flat for some simulated experiments, and why there is a leveling off of the volcano plot (for example, first column, second row). (`viz_expression_data.ipynb`). I will likely move this into an archive folder in a future refactor of the repo
image

BLUE: distribution of adjusted p-values for simulation experiment with flat-top (first row, second column)
ORANGE: distribution of adjusted p-values for simulation experiment with more V-shape (first row, third column)
image

The TOP heatmap shows the expression in the simulated experiment with flat top (BLUE) for those genes with low adjusted p-values (i.e. in the peak of the BLUE distribution). The BOTTOM heatmap shows the expression of the simulated experiment with a more V-shape (ORANGE) for those genes with low adjusted p-values (i.e. those with adjusted p-values = 0-0.4). For the more V-shaped experiment, it looks there was more consistency amongst samples within the group (i.e. WT vs mutant). Depending on the simulation experiment generated, there might be some noise created by the VAE. 

image

image

**Takeaway:**

In trying to understand why there are these flat-tops to some of the volcano plots and why some volcano plots are completely flat, we found:

1. This behavior is not a result of how we are plotting in python (there was some speculation about there being an issue with the numpy library used)
2. The latent space shifting we're doing seems to roughly preserve differences between groups, but this signal can be muddled/noisy depending on where the experiment was shifted to (i.e. the representation that is found in that location can cause the experiment to have a more compressed difference between groups) as seen in the heatmaps. The heatmap of the two simulation experiments shows that some experiments have a more noisey distinction between groups (WT vs mutant) whereas the other simulation experiment has a more distinct difference where the within grouping is cleaner. This definitely points to the need to understand how this simulation process is working and how biology is represented in the latent space. This will definitely be a project for the future. For now we at least have an explanation for why we are observing these shapes in the volcano plots


pseudomonas_analysis/nbconverted/viz_expression_data.py
```diff
@@ -0,0 +1,579 @@
...
+
+sns.distplot(simulated_expression.loc["GSM822708_wtLB_A.CEL"])
+sns.distplot(simulated_expression.loc["GSM822712_delta_cbrBLB_A.CEL"])
+
+
+# In[10]:
+
+
+# Heatmap visualization
+sns.heatmap(simulated_expression.iloc[:,10:20])
```
jjc2718, line 157: How are you choosing the genes to look at here? Are they just random ones?

pseudomonas_analysis/nbconverted/viz_expression_data.py
```diff
@@ -0,0 +1,579 @@
...
+    simulated_DE_stats_dir,
+    project_id,
+    pval_name,
+    logFC_name,
+    num_simulated,
+    5,
+    5,
+    20,
+    15
+)
```
jjc2718, line 119: In addition to what you pointed out about the flattening, it doesn't look like there are very many significant DEGs in the simulated experiments at all (i.e. the y-axes are a bit compressed), at least compared to the template experiment where there are a lot. Are you surprised by this? I can't remember if you pointed this out before (i.e. in your previous PRs).

pseudomonas_analysis/nbconverted/viz_expression_data.py
```diff
@@ -0,0 +1,579 @@
...
+# 
+# **Takeaway:**
+# 
+# In trying to understand why there are these flat-tops to some of the volcano plots and why some volcano plots are completely flat, we found:
+# 1. This behavior is _not_ a result of how we are plotting in python (there was some speculation about there being an issue with the numpy library used)
+# 2. The latent space shifting we're doing seems to roughly preserve differences between groups, but this signal can be muddled/noisy depending on where the experiment was shifted to (i.e. the representation that is found in that location can cause the experiment to have a more compressed difference between groups) as seen in the heatmaps. The heatmap of the two simulation experiments shows that some experiments have a more noisey distinction between groups (WT vs mutant) whereas the other simulation experiment has a more distinct difference where the within grouping is cleaner. This definitely points to the need to understand how this simulation process is working and how biology is represented in the latent space. This will definitely be a project for the future. For now we at least have an explanation for why we are observing these shapes in the volcano plots
```
jjc2718, line 579: >  The latent space shifting we're doing seems to roughly preserve differences between groups

Is this based on the heatmap for the top DE genes? I wonder if there's some way you can quantitatively show the difference (you don't have to answer this question now, just something to think about for the future)

jjc2718 (approved):
Looks good to me! Nice investigation - definitely some interesting questions raised here.

pseudomonas_analysis/nbconverted/viz_expression_data.py
ajlee21: We think this is the result of the variance shrinkage from the VAE. Though I didn't expect the shrinkage to be so large. This is probably because of the original signal being fairly weak and so its possible that the VAE could compress this already weak signal even more.

pseudomonas_analysis/nbconverted/viz_expression_data.py
ajlee21: This is a good point. I will make sure to update the text here. The reason we're claiming that the differences are preserved roughly is from the previous publication showing that heatmaps between simulated and template experiments had a consistent sample structure but there was a different set of related groups of genes that were DE indicating that the simulated experiment represented a new real experiment.

pseudomonas_analysis/nbconverted/viz_expression_data.py
ajlee21: These were just randomly selected to get a little sense for the expression changes

pseudomonas_analysis/nbconverted/viz_expression_data.py
```diff
@@ -0,0 +1,579 @@
...
+    header=0, 
+    index_col=0
+)
+
+
+# In[20]:
+
+
+sns.distplot(leveloff_simulated_DE[pval_name])
+sns.distplot(simulated_DE[pval_name])
```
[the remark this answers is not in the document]
ajlee21: I didn't but would assume that there would just be one peak. And that you would see a similar pattern in the gene expression as with the flat-topped volcano plots

PR #56, opened 2021-01-13
Change normalized data's URL from AWS S3 to Google Storage

dongbohu:
@ajlee21: Try the new URL. When you click it, the download window should show up.


dongbohu:
By the way, since that file is public, you don't need to log into Google Cloud Platform or use Google Cloud SDK command to download it.

ajlee21 (approved):
Looks like its working! Thanks for updating the readme

PR #57, opened 2021-01-22
Pseudomonas validation

ajlee21:
This PR performs a similar validation, previously performed on recount2 data, to pseudomonas data. Here we are comparing the ranking of genes generated by SOPHIE vs those from a manually curated dataset, GAPE.

The following changes were made:
1. Added `0_prepare_reference_gene_file.ipynb` notebook that processes the curated ANOVA results to get gene rankings. These gene rankings will be what we compare our SOPHIE rankings against
2. Update the existing `2_identify_generic_genes_pathways.ipynb` notebook to compare SOPHIE rankings vs the manually curated ones.
3. Based on validation results seen below, we added additional notebooks to test hypothesis (see below). These new notebooks include `0_subset_training_compendium.ipynb` to create new training compendium and `2_identify_generic_genes_pathways_pao1.ipynb` to perform validation analysis on the new training compendium. The code in this notebook is nearly identical to `2_identify_generic_genes_pathways.ipynb` and so doesn't need much review. There were some cutsom edits that needed to be made due to limitations in ponyo. Which I have created an issue for.
4. Update the supporting functions to use a different reference. Previously the functions were written assuming the validation was only performed on the recount2 dataset, so there were things that were hard-coded in.

The main result is here:
image

* Overall there is good consistency between SOPHIE and the reference set of experiments. 
* Genes that SOPHIE says are generic, GAPE doesn't find as strongly generic
* Genes that SOPHIE says are not generic, GAPE also says is not generic (genes that consistently didn’t change or genes that changed in a subset of cases).
* There is some noise in the bottom right corner (i.e. genes that the reference didn't think were as generic but SOPHIE did). These might be the result of the reference being limited to 73 experiments and differences in data processing (ANOVA vs RMA).
* One hypothesis is that the subset of genes that are found to be generic by SOPHIE are not generic using GAPE-curated experiments because SOPHIE is trained on a compendium containing multiple strains, whereas we suspect that the GAPE experiments are only from a single strain (PAO1). So genes that are generic in other strain contexts will not be detected by GAPE. To test this we filtered the SOPHIE training compendium to only include PAO1 strains.

Using only PAO1 samples we get:
image

* Compared to the correlation plot comparing SOPHIE trained on the full pseudomonas compendium vs GAPE experiments [here](2_identify_generic_genes_pathways.ipynb), it looks like there is more noise in the above correlation plot. Looking up the experiment ids associated with GAPE [here](https://github.com/DartmouthStantonLab/GAPE/blob/main/Pa_GSEnum_unzip.rds) it looks like GAPE does contain a mix of PAO1, PA14 and clinical strains. So by subsetting out training compendium to only represent PAO1 patterns, which is adding more noise to the bottom left of the plot as expected (genes that GAPE finds generic are not found by SOPHIE due to our limited representation in our training compendium)

This inconsistency in genes found to be generic by SOPHIE and not by the manually curated set of experiments appears in this analysis using _P. aeruginosa_ data and also human data. Some other hypotheses to test in the future include:
1. For Pseudomonas, perhaps this is a difference in how the data was processed. This would require us to re-process the manually-curated experiments, which would take some time
2. In humans, perhaps this is a difference in the platforms (RNAseq vs array). We would need to find an RNAseq compendium that has DE analysis uniformly processed and available.
3. Perhaps the difference is due to context differences. Thought [this notebook](https://github.com/greenelab/generic-expression-patterns/blob/master/human_cancer_analysis/2_identify_generic_genes_pathways.ipynb) compared genes using SOPHIE trained on cancer-specific contexts vs Crow et. al. using mixed context. And there was noise in the middle but no distinct group on the bottom right.


ajlee21:
Note: commit messages ccf38bb and earlier are from a merge conflict from https://github.com/greenelab/generic-expression-patterns/pull/54 that I addressed but resulted in my ajlee master branch getting out of sync with greenelab master. So these commits are already merged in greenelab master but still showing in the history. These commits do not reflect the current changes made in this PR

generic_expression_patterns_modules/process.py
```diff
@@ -449,12 +429,12 @@ def process_raw_compendium_pseudomonas(
     # Create processed pseudomonas compendium data file
     raw_compendium = pd.read_csv(raw_filename, header=0, index_col=0, sep="\t")
 
-    if raw_compendium.shape != (950, 5549):
+    if raw_compendium.shape[1] != 5549:
```
ben-heil, line 432: Is there a reason that there will always be exactly this many genes/samples?

pseudomonas_analysis/0_prepare_reference_gene_file.ipynb
```diff
@@ -0,0 +1,2456 @@
...
+    "    if i == 0:\n",
+    "        reference_stats_df = reference_df[\"DEG\"].to_frame(\"experiment_0\") \n",
+    "    else:\n",
+    "        reference_stats_df = pd.merge(\n",
+    "            reference_stats_df,\n",
+    "            reference_df[\"DEG\"].to_frame(f\"experiment_{i}\"),\n",
+    "            left_index=True, \n",
+    "            right_index=True,\n",
+    "            how='left'\n",
+    "        )\n",
```
ben-heil, line 2112: If this cell runs slowly, you can probably speed it up by keeping track of the DEGs for each experiments in a dictionary then add create the dataframe at the end

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+    "\n",
+    "    model_encoder_file = glob.glob(os.path.join(NN_dir, \"*_encoder_model.h5\"))[0]\n",
+    "\n",
+    "    weights_encoder_file = glob.glob(os.path.join(NN_dir, \"*_encoder_weights.h5\"))[0]\n",
+    "\n",
+    "    model_decoder_file = glob.glob(os.path.join(NN_dir, \"*_decoder_model.h5\"))[0]\n",
+    "\n",
+    "    weights_decoder_file = glob.glob(os.path.join(NN_dir, \"*_decoder_weights.h5\"))[0]\n",
+    "    \n",
+    "    #print(\"model files used\")\n",
```
ben-heil, line 292: You can probably remove the commented out print statements in this cell

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+    "    NN_architecture,\n",
+    "    dataset_name,\n",
+    "    scaler,\n",
+    "    local_dir,\n",
+    "    base_dir,\n",
+    "    run,\n",
+    "):\n",
+    "    \"\"\"\n",
+    "    Generate new simulated experiment using the selected_experiment_id as a template\n",
+    "    experiment using the same workflow as `simulate_by_latent_transform`\n",
```
ben-heil, line 232: What is `simulate_by_latent_transform`? It may help readers to reference what file it's in

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+    "    Returns\n",
+    "    --------\n",
+    "    simulated_data_file: str\n",
+    "        File containing simulated gene expression data\n",
+    "\n",
+    "    \"\"\"\n",
+    "\n",
+    "    # Files\n",
+    "    NN_dir = os.path.join(base_dir, dataset_name, \"models\", NN_architecture)\n",
+    "    latent_dim = NN_architecture.split(\"_\")[-2] # Here we manually changed the index to work\n",
```
ben-heil, line 282: I don't know how frequently you use the `NN_architecture` string, but you may want to just pass the latent dimension as a parameter rather than having this function's behavior be dependent on the architecture name

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+    "\n",
+    "    # Find a new location in the latent space by sampling from the latent space\n",
+    "    encoded_means = data_encoded_all_df.mean(axis=0)\n",
+    "    encoded_stds = data_encoded_all_df.std(axis=0)\n",
+    "\n",
+    "    latent_dim = int(latent_dim)\n",
+    "    new_centroid = np.zeros(latent_dim)\n",
+    "\n",
+    "    for j in range(latent_dim):\n",
+    "        new_centroid[j] = np.random.normal(encoded_means[j], encoded_stds[j])\n",
```
ben-heil, line 335: Why is the new centroid a random variable rather than being set to the centroid of the template data?

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+       "Rank (Real)                          False\n",
+       "abs(logFC) (Real)                    False\n",
+       "logFC (Real)                         False\n",
+       "Median adj p-value (simulated)       False\n",
+       "Rank (simulated)                     False\n",
+       "Mean abs(logFC) (simulated)          False\n",
+       "Std deviation (simulated)            False\n",
+       "Number of experiments (simulated)    False\n",
+       "Z score                              False\n",
+       "Gene Name                             True\n",
```
ben-heil, line 1088: Is it okay for the Gene Name column to have NaNs?

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Takeaway:**\n",
+    "\n",
+    "* X-axis: gene ranking using SOPHIE (trained on Pseudomonas compendium containing ~1K experiments)\n",
+    "* Y-axis: gene ranking using GAPE (curated set of 73 experiments)\n",
+    "\n",
+    "* Compared to the correlation plot comparing SOPHIE trained on the full pseudomonas compendium vs GAPE experiments [here](2_identify_generic_genes_pathways.ipynb), it looks like there is more noise in the above correlation plot. Looking up the experiment ids associated with GAPE [here](https://github.com/DartmouthStantonLab/GAPE/blob/main/Pa_GSEnum_unzip.rds) it looks like GAPE does contain a mix of PAO1, PA14 and clinical strains. So by subsetting out training compendium to only represent PAO1 patterns, which is adding more noise to the bottom left of the plot as expected (genes that GAPE finds generic are not found by SOPHIE due to our limited representation in our training compendium)"
```
ben-heil, line 1207: It may be helpful to plot the line y=x to show which areas have more GAPE genes or more SOPHIE genes. It may not though, I was just confused that it looked to not be symmetrical at first

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
```diff
@@ -0,0 +1,1764 @@
...
+    "summary_pathway_ranks.to_csv(pathway_summary_filename, sep='\\t')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Quick check:**\n",
+    "\n",
+    "Looks like Ribosomes are found to be significantly differentiated in the template and across 25 simulated experiments. So this pathway looks to be generic, which is consistent with findings from [Crow. et. al.](https://www.pnas.org/content/116/13/6491.abstract), using Human data. This is also consistent with [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2679180/) which observed that Ribosome proteins have many different functions in *S. cerevisiae*, which might explain their variability across a wide range of contexts.\n",
```
ben-heil, line 1737: Is there a reason why all the pathways in this table have the same p-value?

ben-heil (approved):
Looks good! Sorry for leaving so many comments, it's late (for me) so I'm more confused than usual

generic_expression_patterns_modules/process.py
ajlee21: This is the number of genes specific to this pseudomonas compendium, which this function is designed to process data for. I think I will add a comment to explain that though so a user doesn't mistake this for any pseudomonas dataset.

pseudomonas_analysis/0_prepare_reference_gene_file.ipynb
ajlee21: This cell is very quick but thank you for the nice tip. Always helpful to know how to cleanup my code

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Good catch. Just some leftover debugging stuff

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Ah. Since I copied this function, its no longer in the same file as the `simulate_by_latent_transform.py` function that includes a description. I'll update this text

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Ha...you found the source of the problem. For some reason past me thought it made sense to extract the latent dim from the `NN_architecture` string. I have added an issue to ponyo to fix this in a new version: https://github.com/greenelab/ponyo/issues/26

For now I'm doing it this hacky way because this repo is mostly using functionality found in an earlier version of ponyo. So to avoid more things breaking and having to update ponyo and this repo in parallel I've frozen the version of ponyo used here. Not an ideal solution. But something I chose to do to keep the analysis of this repo moving forward in a timely manner.

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Yes this is ok. Some PA### ids don't have a gene name associated.

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: So the idea behind this simulation is that it takes a template experiment (this is your real gene expression experiment) and linearly shifts this template experiment to a new random location (where the new location is based on the latent space distribution). This shifted template experiment becomes your simulated experiment. So the shift vector is connecting the centroid of the template experiment to the random new location (new_centroid). Hopefully that makes more sense.

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Good eye. I did check that not all pathways have the same p-value. Looks like the adjusted-pvalue  is a ratio of two distributions: (1) the actual enrichment score versus the enrichment scores for all gene sets against all permutations of the dataset and (2) the actual enrichment score versus the enrichment scores of all gene sets against the actual dataset. Looking at the distribution of FDR there tends to be a pileup at the extremes which I think explains this result.

pseudomonas_analysis/2_identify_generic_genes_pathways_pao1.ipynb
ajlee21: Let me make of note of this and think about it some more. It might make sense to do.

PR #60, opened 2021-02-06
Add additional stats columns and redo analysis

ajlee21:
This PR is adding two additional columns to the summary statistics 1) Percentile rank score per gene/pathway and 2) Frequency that gene is found to be DE across simulated experiments

Many files were changed since I needed to rerun the analyses after adding this additional functionality. But these files were previously reviewed in previous PRs and haven't changed other than being rerun. The only main change that needs to be reviewed can be found in `ranking.py`

Note: there is some temporary issue with the pytest server but I have checked that these test run successfully offline


danich1 (approved):
LGTM! I do recommend implementing a linter as a github's pre-commit hook. By using a pre-commit hook it formats your python code before allowing you to push commits. This would fix most of the stylistic comments I would have made. If interested feel free to let me know I can help install it for you.

PR #61, opened 2021-02-09
Update comparison of gene ranks across experiments

ajlee21:
This PR compares gene ranking generated by running SOPHIE 2 times using the same template experiment but different seeds.

Main changes can be found in `compare_generic_genes.ipynb`

Minor change that doesn't need review: update pre-commit files

The main result is:
image

* Running SOPHIE twice using the same template experiment will generate 2 different sets of simulated experiments.
* Since the template experiment is the same, these 2 sets of simulated experiments will have the same experimental design structure/biological context
* As expected, the concordance is very high especially for high ranked and low ranked genes. The genes in the middle rank are more sensitive to changes so you don't get as clear of a signal compared to the extreme ranked genes.

By comparison, if we run SOPHIE using two different template experiments we get,
image

* Compared to the correlation plot above, there are genes in the off-diagonal regions that might indicate that there are generic within the given context of the specific experiment.

Note: there is some temporary issue with the pytest server but I have checked that these test run successfully offline


compare_experiments/compare_generic_genes.py
```diff
@@ -0,0 +1,324 @@
...
+        ]
+    )
+    / len(template2_mean[discordant_genes]),
+)
+
+f4 = sns.distplot(template_2.mean()[discordant_genes], kde=False)
+f4.set_title(f"Expression of discordant genes in {project_id2}")
+f4.set_xlabel("log(gene expression)")
+f4.set_ylabel("log(count)")
+f4.set(xscale="log", yscale="log")
```
ben-heil, line 313: Are the boxes different sizes because this is a log-log plot? Or is there just a consistant number of genes in the low-expression region?

compare_experiments/compare_generic_genes.py
```diff
@@ -0,0 +1,324 @@
...
+    transparent=True,
+    pad_inches=0,
+    dpi=300,
+)
+# -
+
+# **Takeaway:**
+#
+# * Looks like there is good concordance between highly ranked genes (i.e. generic genes)
+# * By comparison if we run SOPHIE using two different template experiments, there are genes in the off-diagonal regions that might indicate that there are generic within the given context of the specific experiment.
```
ben-heil, line 182: What is the difference between a gene that is generic within the context of a specific experiment and a gene that is just upregulated within that experiment? Is it generic in that it appears across several seeds?

ben-heil (approved):
Awesome! Just left a few conceptual questions

compare_experiments/compare_generic_genes.py
ajlee21: Generic genes are those that are DE, differentially up or down regulated across multiple experiments/seeds. So the idea is that some gene experiment designs or biological contexts, could in theory yield different generic genes (i.e. genes that are frequently DE in a cancer context are not necessarily the same genes that are frequently DE in the context of some other stimuli)

compare_experiments/compare_generic_genes.py
ajlee21: Looks like its because of the scale

PR #62, opened 2021-02-09
Update enrichment method to use array data and make new figure

ajlee21:
This PR updates the enrichment analysis. Specifically,

1. Creates figures `enrichment_paired_plot_array.svg` and `enrichment_paired_plot_rnaseq.svg`

These figures show _____

2. Removes ROAST method, which returns unexpected results. In order to debug this, we would need to do further digging into the math behind the method. However, given our time constraint with the manuscript we are removing this method.
3. Update comments in the notebook to more carefully describe this analysis

Note: nbconverted scripts are currently out of sync in this PR but will be updated once precommit changes are merged in.


jjc2718 (approved):
Looks good! Some general questions/comments:

* In the paired plots (both for RNA-seq and array data), do you have any idea why GSVA seems to be inverted (lower genericness = lower enrichment score) compared to the other methods (higher genericness = lower p-value)? Is this just because low p-values correspond to high enrichment, so the correlations are really all in the same direction? If this is true, maybe you could plot -log(p-value) instead of raw p-values so that high x-axis values always correspond to high enrichment (or alternately, 1 - enrichment instead of enrichment for GSVA => low x-axis values == high enrichment).
* I wonder why there seems to be a correlation between GSEA and GSVA for the array data, but not so much for the RNA-seq data. Is there anything that varies between your pipelines for the two data types that might cause this difference?

ajlee21:
> * In the paired plots (both for RNA-seq and array data), do you have any idea why GSVA seems to be inverted (lower genericness = lower enrichment score) compared to the other methods (higher genericness = lower p-value)? Is this just because low p-values correspond to high enrichment, so the correlations are really all in the same direction? If this is true, maybe you could plot -log(p-value) instead of raw p-values so that high x-axis values always correspond to high enrichment (or alternately, 1 - enrichment instead of enrichment for GSVA => low x-axis values == high enrichment).

You're correct, high enrichment scores and low p-values correspond to those genes that are generic. Its probably more intuitive to see a positive correlation along the diagonal of these pair plots so I can update the plotting as you suggested to help with interpretation later.

> * I wonder why there seems to be a correlation between GSEA and GSVA for the array data, but not so much for the RNA-seq data. Is there anything that varies between your pipelines for the two data types that might cause this difference?

Good question. The array data is using a cancer line specific dataset while the RNA-seq dataset is using a more general compendium (recount2). Given that the array dataset is more uniform and GSVA method is sample-wise, which can be more sensitive to noise this would be my guess as to why there is less consistency between the two in the RNA-seq dataset. Good thing to note.

PR #63, opened 2021-02-11
Add scaler transform files to repo

ajlee21:
This PR is adding scaler transform files used to normalize the input data to the repo and updates the config file with the corresponding location


ben-heil (approved):
Thanks for adding these!

PR #64, opened 2021-02-13
Train multiplier on pseudomonas dataset

ajlee21:
This PR follows up on https://github.com/greenelab/generic-expression-patterns/pull/58

This previous PR found different trends using human data vs using pseudomonas data. Since the LV coverage analysis used two different models (multiPLIER for human and eADAGE for pseudomonas), in order to correct for the effect of using different models, here we train PLIER on pseudomonas dataset and re-runs LV coverage analysis. Now when we compare the trends in generic gene coverage between human vs pseudomonas data, where the LVs were both generated using the same PLIER model. 

Coverage of high weight generic genes in humans is: 
image

* In humans, other genes are highly weighted in more LVs compared to generic genes. This would indicate that generic genes contribute alot to few LVs. Perhaps these generic genes are related to a few hyper-responsive pathways. By quick visual (`Generic_LV_summary_table.tsv`) inspection, it looks like many LVs are associated with immune response, signaling and metabolism. Which are consistent with the hypothesis that these generic genes are related to hyper-responsive pathways.

Coverage of high weight generic genes in Pseudomonas is:

image

* In _P. aeruginosa_ both generic and other genes are highly weighted in a similar number of LVs, but overall generic genes contribute a lot to very few LVs. The only associated LV is related to type IV secretion system, which is a complex responsible for a broad range of functions: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3070162/ 

Overall, using human and _P. aeruginosa_ data find a similar proportion of LV (~2%) that are generic genes contribute highly to. But there is a difference in trend between generic vs other genes still.  One possible explanation for the difference is that these generic genes found in human could be the result of platform differences between experiments and there is no LV in PLIER that will represent this so the number of LVs should be higher for generic genes. Perhaps this difference is the results of humans being multicellular and many generic genes being found due to differences in cell type which doesn't apply to bacteria. 

Additional experiments will need to be performed to try to determine the difference. Now that we've corrected for the model used, we might also want to correct for the pathway annotations used to try to make sure the types of processes represented are similar. 

The main changes are:
1. `0_process_Pa_KEGG_pathways.ipynb`: notebook to format KEGG annotations to input into PLIER
2.  `train_PLIER_pa.R`: script to run PLIER on Pa data
3.  `0_process_Pa_multiplier_model.ipynb` notebook to process outputs from PLIER (You have already reviewed code similar to this)
4.  `1_get_multiplier_LV_coverage_Pa.ipynb` notebook to perform LV coverage analysis using Pa PLIER model (You have already reviewed code similar to this)


LV_analysis/1_get_multiplier_LV_coverage_Pa.py
```diff
@@ -0,0 +1,335 @@
...
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:generic_expression] *
+#     language: python
+#     name: conda-env-generic_expression-py
+# ---
+
+# # Coverage of MultiPLIER LV
```
jjc2718, line 16: Do you want to note here that this script is for Pseudomonas data? Looks like this intro is just copy-pasted from the human version of the script (which is fine if that's what you want).

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
```diff
@@ -0,0 +1,335 @@
...
+config_filename = os.path.abspath(
+    os.path.join(base_dir, "configs", "config_pseudomonas_33245.tsv")
+)
+
+params = utils.read_config(config_filename)
+
+local_dir = params["local_dir"]
+
+project_id = params["project_id"]
+quantile_threshold = 0.97
```
jjc2718, line 64: Looks like this parameter is slightly different than the human script (0.97 vs. 0.98) - is that intentional? Same question for the percentile used to define generic genes (60 in human script vs. 80 in this script).

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
```diff
@@ -0,0 +1,335 @@
...
+].values
+generic_highweight = all_coverage_df[all_coverage_df["gene type"] == "generic"][
+    "highweight LV coverage"
+].values
+
+(stats, pvalue) = scipy.stats.ttest_ind(other_highweight, generic_highweight)
+print(pvalue)
+
+# +
+# Check that coverage of other and generic genes across all LVs is NOT signficantly different
```
jjc2718, line 263: So these _are_ significantly different here, right, since the p-value is small? Meaning generic genes on average contribute to fewer LVs than specific genes do?

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
```diff
@@ -0,0 +1,335 @@
...
+    format="svg",
+    bbox_inches="tight",
+    transparent=True,
+    pad_inches=0,
+    dpi=300,
+)
+# -
+
+# **Takeaway:**
+# * In the first nonzero boxplot, generic and other genes are present in a similar number of LVs. This isn't surprising since the number of genes that contribute to each LV is <1000.
```
jjc2718, line 332: This isn't what the statistical test shows, is it (see previous comment)? I guess if you don't believe the results of the t-test that's valid, the medians on the box plot don't look terribly different at least by eye, but it might be good to say that explicitly (unless I'm mistaken about what you're claiming here).

jjc2718 (approved):
Code looks good! My main conceptual question is about your interpretation of the t-test results in the Pa MultiPLIER notebook, and I also added a few other small comments.

LV_analysis/train_PLIER_pa.R
```diff
@@ -0,0 +1,57 @@
...
+    )
+)
+
+# Location of PLIER training script
+source("../generic_expression_patterns/plier_util.R")
+
+# run PLIER model
+plier.result <- PLIERNewData(expression_data)
+
+# TO DO: Save locally
```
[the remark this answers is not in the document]
ajlee21: yup leftover comment. Thanks for checking

generic_expression_patterns_modules/plier_util.R
```diff
@@ -0,0 +1,51 @@
+# Imported from https://github.com/greenelab/multi-plier/blob/master/util/plier_util.R
+PLIERNewData <- function(exprs.mat, seed = 12345) {
+  # A wrapper function for applying PLIER to a data set. We use the following
```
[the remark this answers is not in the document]
ajlee21: Yes this is just copy and pasted from the original repo. I will update this though for my case

LV_analysis/0_process_Pa_KEGG_pathways.py
```diff
@@ -0,0 +1,86 @@
+# ---
```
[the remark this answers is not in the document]
ajlee21: Since I'm now using the pre-commit I have some duplicated files. I plan to refactor this repo after all the analyses are complete to either move all the .py files to a new directory or remove the nbconverted dir. I haven't decided just yet so I'll leave this for that future refactor PR.

LV_analysis/1_get_eADAGE_LV_coverage.py
```diff
@@ -0,0 +1,328 @@
...
+# +
+# Load gene summary data
+data = pd.read_csv(summary_data_filename, sep="\t", index_col=0, header=0)
+
+# Check that genes are unique since we will be using them as dictionary keys below
+assert data.shape[0] == len(data["Gene ID"].unique())
+# -
+
+# **Manual steps to process eADAGE data**
+# 1. Data downloaded from https://zenodo.org/record/580093
```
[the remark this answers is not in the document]
ajlee21: Thanks for the tip! I'll definitely keep this in mind for the future. This is something that just needs to be run once but in general I still like the idea of having a script to do this to try to make things more reproducible as possible. Since we're trying to wrap up the analyses for this manuscript so I don't think I'll add this for this repo but will keep it in mind for the future. Thank you!

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
ajlee21: Thanks for checking. This was intentional. I was trying to vary the quantile to get similar numbers of high weight genes between Pa and human in case that contributed a difference in trend. The generic threshold is based on the correlation plots and so this varied based on the dataset too.

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
ajlee21: Yes. you're interpretation is correct here. More below

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
ajlee21: Yes I will clarify this comment. You're correct that despite the t-test saying there is a significant difference, the distribution looks very similar. So I wasn't planning to claim a difference despite the t-test statistic.

LV_analysis/0_process_Pa_KEGG_pathways.py
```diff
@@ -0,0 +1,86 @@
...
+# # Process _P. aeruginosa_ KEGG pathways
+#
+# This notebook downloads _P. aeruginosa_ KEGG pathways from [here](https://raw.githubusercontent.com/greenelab/adage/master/Node_interpretation/pseudomonas_KEGG_terms.txt). These are the pathways used in the [ADAGE paper](https://msystems.asm.org/content/1/1/e00025-15) and formats it to be input to PLIER. PLIER expects pathway data to be a matrix of the form gene x pathway, where the values are 1 if the gene is contained with the pathway and 0 otherwise.
+#
+# This formated pathway data is used as in [PLIER R script](../generic_expression_patterns_modules/plier_util.R).
+
+import pandas as pd
+
+# Load data
+expression_filename = "https://raw.githubusercontent.com/greenelab/adage/master/Data_collection_processing/Pa_compendium_02.22.2014.pcl"
```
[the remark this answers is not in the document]
ajlee21: Thank you for the great suggestion. I did not think about this

LV_analysis/1_get_multiplier_LV_coverage_Pa.py
jjc2718: Yeah - I think the sample size is fairly high (~10,000 genes, I'm guessing?) so even a small difference will be significant. That doesn't necessarily invalidate the test, but it seems valid to argue that the difference is small enough that it is unlikely to matter biologically.

[This stackexchange thread](https://stats.stackexchange.com/a/2519) might be interesting, along those lines.

ajlee21: Thank you for sharing this. This is very interesting. I think I'll need to think about this some more, its not really intuitive to me why increasing sample size might lead to increased chance of significance even for a small effect. I guess my understanding is that the t-test is comparing means between two distributions, so I'm not sure why increasing sample size might yield this

PR #65, opened 2021-02-17
Create G-G network visualization

ajlee21:
This PR is adding an R script to plot G-G network to visualize the distribution of generic genes.

Changes in this PR include:
1. `pseudomonas_analysis/prepare_data_for_GiG_network.ipynb` notebook to process data needed to input into visualization script
2. `generic_expression_patterns_modules/make_GiG_network.R` R script to visualize G-G network where each node is a gene, edge weights are based on the correlation betwween genes using eADAGE LV and generic genes are highlighted in red.

The G-G network can be found in slide 2 here: https://docs.google.com/presentation/d/1rU70mCkcRPigrhpMQMgbYR9hXq2yzE9zD-LIuaVXWSc/edit?usp=sharing

Based my meeting, depending on the text we have in the manuscript, this network may be removed or changed.


generic_expression_patterns_modules/make_GiG_network.R
```diff
@@ -0,0 +1,41 @@
+# Using R 3.6.3
+
+# Manually load operons.rda and geneinfo.rda into R environment
+# from https://github.com/greenelab/ADAGEpath/blob/master/data/
+if (!requireNamespace("BiocManager", quietly = TRUE))
+  install.packages("BiocManager")
+BiocManager::install("impute")
```
jjc2718, line 7: Do these need to be installed every time you run the script? Maybe you could check if they're installed before installing them. Something like this might work: https://stackoverflow.com/a/4090208

generic_expression_patterns_modules/make_GiG_network.R
```diff
@@ -0,0 +1,41 @@
...
+# from https://github.com/greenelab/ADAGEpath/blob/master/data/
+if (!requireNamespace("BiocManager", quietly = TRUE))
+  install.packages("BiocManager")
+BiocManager::install("impute")
+BiocManager::install("greenelab/ADAGEpath")
+
+# Get eADAGE model
+eADAGEmodel <- ADAGEpath::eADAGEmodel
+
+# Get all eADAGE signatures
```
jjc2718, line 13: I'm not too familiar with how eADAGE/ADAGEpath plots networks. Are you only generating a network for certain genes (or certain pathways/gene signatures), or is this all Pseudomonas genes/interactions? 

What do the edges represent here: pathway relationships? Physical interactions between genes/proteins?

I know a lot of these questions are probably answered in the ADAGEpath repo, but it might be good to give a short synopsis here.

jjc2718 (approved):
Code looks good to me!

I guess my main conceptual question would be about your interpretation of this network plot. Is the takeaway here that generic genes tend to cluster in similar network regions (i.e. similar pathways or functional processes)? If so, I may have some *opinions* about that but I'll spare you for now, until you tell me how you're interpreting it.

I assume you've already talked through this with Casey and your collaborators, so don't worry about answering in too much detail if you're not planning to use this network much going forward.

generic_expression_patterns_modules/make_GiG_network.R
ajlee21: This is a very good point. I will add this detail to the script. Essentially the networks were generated by using weights generated by eADAGE (denoising autoencoder) where nodes = genes and edges = similar weight profiles for how much that gene contributes to the eADAGE LV. This network is showing all genes, not a subset. The clustering is using pearson correlation with a default cutoff of 0.5

ajlee21:
Very good question. For this PR we were hoping to say something more about the biology of these generic genes found. Here we were curious to determine if there were certain pathways in which generic genes clustered. The the number of genes and connections is a bit difficult for us to tell what is going on here. Will followup with you offline for suggestions on how we can improve this analysis

PR #66, opened 2021-02-18
Add new volcano plots

ajlee21:
There isn't much code to review in this PR. This PR is adding a notebook to plot 2 new volcano plots that correspond to the wet lab experiments performed by the Hogan lab.

The experiment is comparing WT vs cbrB mutant strains. Using out SOPHIE method, we identified candidate specific genes (red) that the Hogan lab nicely validated were specific to this cbrB context.

The volcano plots can be found in `viz_E-GEOD-33245_cbrB_v_WT.ipynb `


arielah:
This looks good! My only suggestion is to add a little bit of extra information at the top of the two visualization scripts. I'm not sure what the difference between viz_expression_data and viz_E-GEOD-33245_cbrB_v_WT is or why they're two separate scripts, since the code looks very similar.

Also, I didn't dig too far into the checks failing, but file not found errors are usually a quick fix. Hope that's the case here!
FileNotFoundError: File b'data/input/mapped_recount2_template_data.tsv' does not exist

arielah (approved):

ajlee21:
> This looks good! My only suggestion is to add a little bit of extra information at the top of the two visualization scripts. I'm not sure what the difference between viz_expression_data and viz_E-GEOD-33245_cbrB_v_WT is or why they're two separate scripts, since the code looks very similar.

This is a very good point. One of the visualizations is more general then the other one is the main difference. The other is specific for one experiment.

> Also, I didn't dig too far into the checks failing, but file not found errors are usually a quick fix. Hope that's the case here!
> FileNotFoundError: File b'data/input/mapped_recount2_template_data.tsv' does not exist

This is a pytest that is failing for reasons that I'm not sure because pytest passes when I run it locally. And the file it says its missing is there.

PR #67, opened 2021-02-25
Update figures

ajlee21:
This PR is updating figures for the manuscript. Not much code to review here, but please let me know if you have thoughts on the figures (though it might be easier when you can review it with the text)


jjc2718 (approved):
Code looks good to me!

Just curious - what's your interpretation of the new figures you added in Figure 1D (i.e. in `compare_experiments/concordance_between_diff_recount2_templates.svg` and `compare_experiments/concordance_between_same_recount2_templates.svg`)? Are you showing that simulating data twice with the same experiment results in a ranking (or set of generic genes) that is more consistent than if you simulate data with different experiments from recount2?

I agree this will probably be easier to see with the rest of the manuscript, just wondering why you decided to add this part of the figure.

ajlee21:
Good question. So my takeaway for these additional panels would be that having shown that SOPHIE’s generic percentiles ranking can recapitulate percentiles of genes using a manually curated dataset, here we examined the robustness of these generic patterns. 

So we compared SOPHIE percentiles using the same template experiment and showed that we get good concordance especially for high and low percentile genes (Figure 2E). The genes in the middle percentile are more sensitive to changes so we don’t get a clear signal. This noise is even more pronounced when we compare the percentiles generated using two different template experiments (Figure 2F). However, overall, we get good consistency of generic gene percentiles across different template experiments

ajlee21:
Still not sure why pytest is failing here because it runs locally ok. Will add an issue and return to this

PR #68, opened 2021-03-01
Community detection analysis on eADAGE correlation network

jjc2718:
This analysis explores the position of generic genes in a correlation network between Pa genes derived from eADAGE weights (see README for additional details).

The main goal of the analysis was to test the hypothesis that generic genes would be spread across fewer network communities than random genes in the network with similar degree to the generic genes. I tried two community detection methods (planted partitions and Louvain clustering), and in both cases this hypothesis does seem to hold. For planted partition model:

image

And for Louvain:

image

This seems (to me) to suggest that generic genes may preferentially cluster in a few pathways or functional gene groups, although I'm curious if this matches your interpretation or if there are additional analyses we can do to follow up on this.

I'm still planning to look at some individual node attributes (average degree, betweenness centrality, etc), but I'll make a separate PR for that analysis.


jjc2718:
@ajlee21 Looks like I can't assign a reviewer to PRs in your repo, but if you could take a look at this when you get a chance that would be great!

network_analysis/nbconverted/1_preprocess_network_data.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Preprocess and validate eADAGE network data
+# 
+# eADAGE G-G network/edge list was generated by Alex using the script at `../generic_expression_patterns_modules/make_GiG_network.R`.
```
ajlee21, line 6:
```suggestion
# eADAGE G-G network/edge list was generated by Alex using the script at [make_GiG_network.R](../generic_expression_patterns_modules/make_GiG_network.R). 
```

network_analysis/nbconverted/1_preprocess_network_data.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Preprocess and validate eADAGE network data
+# 
+# eADAGE G-G network/edge list was generated by Alex using the script at `../generic_expression_patterns_modules/make_GiG_network.R`. 
+# 
+# Here, we preprocess it to reweight negative edges, add generic gene info from `../pseudomonas_analysis/annot_df.tsv`, and save in a file that can be quickly loaded by `graph-tool` for subsequent analyses.
```
ajlee21, line 8:
```suggestion
# Here, we preprocess it to 
* reweight negative edges
* add generic gene info from [here](../pseudomonas_analysis/annot_df.tsv)

Output: New edge list file that can be quickly loaded by `graph-tool` for subsequent analyses.
```

network_analysis/nbconverted/1_preprocess_network_data.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+# save graph with attributes to file
+G.save(processed_graph, fmt='gt')
+
+
+# In[10]:
+
+
+# plot generic and non-generic genes, just for fun
```
ajlee21, line 119: Nice! This looks like what I found using the ADAGEpath tool!

network_analysis/nbconverted/1_preprocess_network_data.py
```diff
@@ -0,0 +1,126 @@
...
+processed_graph = os.path.join(data_dir, 'eadage_generic_graph_unsigned.gt')
+
+
+# #### Load edge list and handle negative edge weights
+# 
+# Most algorithms for community detection and betweenness don't work with negative weights/correlations. Taking inspiration from WGCNA, we can solve this problem in one of two ways:
+# * Unsigned: `weight = abs(corr(g1, g2))`
+# * Signed: `weight = abs((1 + corr(g1, g2) / 2)`
+#  
+# We should probably try both eventually, but for now we'll just use the unsigned version (taking absolute value of negative edges). I tend to agree with the rationale in this BioStars post for taking absolute values (unsigned) rather than rescaling (signed approach): https://www.biostars.org/p/144078/#144088
```
ajlee21, line 45: I agree. In this case I think the magnitude is what matters in terms of defining the connection

network_analysis/nbconverted/1_preprocess_network_data.py
```diff
@@ -0,0 +1,126 @@
...
+    edgelist_df['weight'] = edgelist_df.weight.abs()
+    edgelist_df.to_csv(processed_edge_list, columns=['from', 'to', 'weight'], index=False)
+
+edgelist_df = pd.read_csv(processed_edge_list)
+edgelist_df.head()
+
+
+# In[4]:
+
+
```
ajlee21, line 62:
```suggestion
# 1 denotes a generic gene and 0 is all other genes. A gene is considered generic if it had a high percentile from SOPHIE and the manually curated set based on the correlation plot seen [here](../pseudomonas_analysis/2_identify_generic_genes_pathways.ipynb)
```

network_analysis/nbconverted/2A_community_detection_planted_partition.py
```diff
@@ -0,0 +1,160 @@
...
+print(G)
+print(list(G.vp.keys()))
+print(list(G.ep.keys()))
+
+
+# In[4]:
+
+
+# community detection using planted partition model
+state = gt.PPBlockState(G)
```
ajlee21, line 57: Does state include a membership assignment mapping?

network_analysis/nbconverted/2A_community_detection_planted_partition.py
```diff
@@ -0,0 +1,160 @@
...
+
+labels_df = graph_state_to_df(G, state)
+print(len(labels_df.label.unique()))
+labels_df.sort_values(by='degree', ascending=False).head()
+
+
+# In[8]:
+
+
+# simultaneously sort nodes and degrees by degree, ascending
```
ajlee21, line 98:
```suggestion
# simultaneously sort nodes and degrees by degree, descending
```

network_analysis/nbconverted/2A_community_detection_planted_partition.py
```diff
@@ -0,0 +1,160 @@
...
+    sampled = [1 if gene in s_nodes else 0 for gene in labels_df.gene]
+    labels_df['sampled_{}'.format(it)] = sampled
+
+labels_df.sort_values(by='degree', ascending=False).iloc[:5, :5]
+
+
+# In[9]:
+
+
+generic_count_df = (
```
ajlee21, line 116: Do you mind adding comments to these next 2 cells...i got a little lost in the code. It looks like you're grouping by module? but then i'm not sure how you got the additional sampled columns..

network_analysis/nbconverted/2B_community_detection_louvain.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+annot_df = pd.read_csv(generic_gene_map, sep='\t', index_col=0)
+annot_df.head()
+
+
+# In[7]:
+
+
+is_generic = [int(annot_df.loc[v['name'], 'label']) for v in G.vs]
```
ajlee21, line 79: What are you doing here?

network_analysis/nbconverted/2B_community_detection_louvain.py
```diff
@@ -0,0 +1,173 @@
...
+
+# In[13]:
+
+
+n_generic_groups = nonzero_counts_df.iloc[0, 0]
+n_total_groups = len(labels_df.label.unique())
+
+sns.set({'figure.figsize': (8, 5)})
+
+sns.histplot(nonzero_counts_df.iloc[0, 1:].values,
```
ajlee21, line 157: Just curious, is there a reason why the Louvain produces more modules?

ajlee21 (approved):
This is AWESOME!

In our previous results we found that with humans using PLIER LVs, generic genes were found to be high weight in fewer LVs compared to other genes. However, performing the same analysis for pseudomonas  using eADAGE LVs we found a similar coverage of LVs between high weight generic vs other genes. 

For this network analysis we're considering all generic genes and it appears that they are clustered in a few pathways where perhaps some of the generic gene clusters are defined by a high weight generic gene and the number of other gene clusters is the same for this condition...does that maybe make sense?

network_analysis/nbconverted/1_preprocess_network_data.py
jjc2718: Yeah - I don't think it will matter much in either case, since there are very few negative edges in this network, but I do think the magnitude is more important than the direction for finding modules/"pathways".

network_analysis/nbconverted/2A_community_detection_planted_partition.py
jjc2718: I think these are ascending? It actually doesn't matter which direction as long as they're sorted, but the check [here](https://github.com/greenelab/generic-expression-patterns/blob/e67bbc9eecfa0c51a2389f8db7ad764a0e0d7568/network_analysis/sample_nodes.py#L75) is making sure the degrees are sorted from small to large (i.e. ascending, I think).

When I print them below I'm printing in descending order, so maybe that's what's confusing you. I'll change that to ascending order to match the sorting.

ajlee21: Ah yup i was going off the print statement:)

network_analysis/nbconverted/2A_community_detection_planted_partition.py
jjc2718: It does, once the model is fit (which is what the line below this does). See [here](https://graph-tool.skewed.de/static/doc/inference.html#graph_tool.inference.blockmodel.BlockState) for more details (the "block state" is the membership assignments).

network_analysis/nbconverted/2B_community_detection_louvain.py
jjc2718: This just reads the generic/not generic labels from the tsv file you pointed me to into a list, in the same order as the nodes (genes) in the graph. I'll add a comment saying this.

network_analysis/nbconverted/2B_community_detection_louvain.py
jjc2718: My guess would just be overfitting; i.e. many of the small communities should really be a single larger community. [Figure 1 of this paper](https://journals.aps.org/prresearch/pdf/10.1103/PhysRevResearch.2.043271) shows that methods like Louvain (the "Q=" methods) tend to find more communities than the true number of communities in generated graphs.

People have a lot of strong opinions about this sort of thing, though. I picked Louvain because empirically it seems to do pretty well in biological networks: e.g. in [our DREAM challenge paper](https://doi.org/10.1038/s41592-019-0509-5) many of the top performing methods used some variation of Louvain. It's also fairly straightforward to understand. But I also think the more statistically principled models, like the planted partition model described in the first paper I linked above, have some advantages.

In any case, the results here don't seem to be sensitive to the choice of algorithm, which makes me more confident that what we're seeing is a real signal.

jjc2718:
> In our previous results we found that with humans using PLIER LVs, generic genes were found to be high weight in fewer LVs compared to other genes. However, performing the same analysis for pseudomonas using eADAGE LVs we found a similar coverage of LVs between high weight generic vs other genes.

Yeah - I think it's interesting that the eADAGE LV analysis gave a different result, but thinking it through it seems like this is a pretty distinct analysis so it's not too surprising (to me, anyway) that the results are different.

Here, in the eADAGE correlation network, we're grouping genes that have similar signal across all (or most) LVs. We can sort of interpret these groups as the "strongest" pathway or function signal for each gene, rather than the total number of pathways or functions (or other LVs) that each gene is involved in.

So what you said here:
 
> For this network analysis we're considering all generic genes and it appears that they are clustered in a few pathways

seems to me like a valid interpretation of this analysis, but not so much for the LV analysis (that looks at enrichment across many pathways), at least in my view.

This part:

> where perhaps some of the generic gene clusters are defined by a high weight generic gene and the number of other gene clusters is the same for this condition...does that maybe make sense?

I'm not sure I quite understand. You mean that some of the clusters with generic genes might be driven by a single gene, and others might be driven by many generic genes? We could look at that in more detail if it would be interesting, but I want to make sure I get what you're saying first?

ajlee21:
I agree that this analysis is distinct from the LV analysis. I think what confuses me a bit here is how to interpret the modules. So generic genes appear to cluster into fewer modules compared to other genes, where these modules represent some shared LV signal. I guess I normally think of these LV as corresponding to some pathway or function so its hard for me to figure out what these modules are, maybe some broader group of pathways or functions?

The last point was more me trying to think through if there was a way to compare these network results with what we did with the LVs to check consistency. But as we mentioned the analyses are distinct enough that I don't think we can. I was trying to figure out if there was a way to layer the weights of the genes, but you would need to do this for LVs individually so I don't think this would work.

jjc2718:
> I guess I normally think of these LV as corresponding to some pathway or function so its hard for me to figure out what these modules are, maybe some broader group of pathways or functions?

Here's an idea for an analysis we could do here:

* In addition to just testing the number of communities that contain generic or non-generic genes, we could identify which communities contain significantly more generic vs. non-generic genes. The basic idea would be to use the distributions of sampled non-generic gene counts in each community to define a null distribution, then see how far away from the mean the generic gene count is.
* For communities where the generic gene count is significantly higher than any of the sampled non-generic gene counts, we could run some kind of functional enrichment test (using e.g. KEGG pathways).

The result of this would be which communities tend to contain more generic genes than non-generic genes, and roughly what we think they represent in terms of functions or pathways.

I know you're trying to get this paper submitted soon, though, so it's up to you and Casey whether or not you want to do this. I can probably do it within the next week or so, but it'll take a bit of additional work.

> The last point was more me trying to think through if there was a way to compare these network results with what we did with the LVs to check consistency. But as we mentioned the analyses are distinct enough that I don't think we can. I was trying to figure out if there was a way to layer the weights of the genes, but you would need to do this for LVs individually so I don't think this would work.

Yeah, I'm not sure how this would work exactly. I'll think about it, but I do think the interpretation of the analyses is pretty distinct.

cgreene:
I think in Deb's experience eADAGE nodes weren't always cleanly one function and functions were spread across nodes with additional other components. If we were going to go down a path, characterizing betweenness / centrality / degree of generic genes might be the one to go down.

ajlee21:
Thanks @jjc2718 and @cgreene 

Jake I think that the new analysis your proposing will be more equivalent to what we have already done with looking at the enrichment in LVs, which is not to say its a bad analysis to do. But I think the one Casey is suggesting might be more complementary to the LV analysis that we've already done.

In that case I'll go ahead and merge this PR

PR #70, opened 2021-03-03
Compare degree + centrality measures for generic vs. non-generic genes

jjc2718:
Following up on #68, we also wanted to look at some node/edge attributes in the eADAGE weight correlation network, to see if there are any systematic differences between generic and non-generic genes.

I looked at:
* degree distribution
* edge weight distribution
* betweenness centrality
* PageRank

On average, genes tend to have higher degrees in the network than non-generic genes:

image

They also tend to have higher PageRank values on average (somewhat correlated with degree, but not perfectly):

image

Both these quantities tend to suggest (at least to me) that generic genes have a tendency to be gene "hubs" in the network, connected to many other genes. Edge weights and betweenness centrality values, however, don't seem to differ much between generic and non-generic genes.


network_analysis/nbconverted/2_calculate_node_statistics.py
```diff
@@ -0,0 +1,182 @@
...
+# make sure vertex/edge properties exist
+print(G)
+print(list(G.vp.keys()))
+print(list(G.ep.keys()))
+
+
+# In[4]:
+
+
+# plot degree distribution + look at degrees of generic genes
```
ajlee21, line 49: SUPER nitpicky, but i would make this a markdown cell with this comment this this comment extends to multiple cells

network_analysis/nbconverted/2_calculate_node_statistics.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+sns.set({'figure.figsize': (12, 4)})
+sns.set_style('whitegrid')
+fig, axarr = plt.subplots(1, 2)
+sns.histplot(data=weight_df, x='weight', hue='is_generic', element='step',
+             stat='probability', common_norm=False, ax=axarr[0])
+axarr[0].set_title('Weight distribution of generic/non-generic edges')
+sns.boxplot(data=weight_df, y='weight', x='is_generic', ax=axarr[1])
+axarr[1].set_title('Weight distribution of generic/non-generic edges')
```
ajlee21, line 104: So i think that bump means that there are some generic genes that are more likely to be high weight with other generic genes than any other gene..

network_analysis/nbconverted/2_calculate_node_statistics.py
```diff
@@ -0,0 +1,182 @@
...
+# In[8]:
+
+
+# add distance = 1 - edge weight to graph as edge property
+#
+# betweenness centrality is evaluated based on shortest paths in the network,
+# but since our edge weights represent similarity, we need to convert them to
+# a distance/cost measure
+# 
+# taking distance = (1 - similarity) is a simple way to do this
```
ajlee21, line 116:
```suggestion
# taking distance = (1 - similarity) is a simple way to do this
# So a high similarity between genes will have a short distance
```

ajlee21 (approved):
Nice!

This is super interesting...its making me curious if we'd see the opposite for humans since our LV analysis and correlation results suggest that the human generic genes are not hubs but rather concentrated in highly sensitive pathways. I don't think its necessary to have for the paper but would be interesting to see if that is the case.

Not sure if @cgreene has other thoughts

cgreene:
If we're going to look across human and pseudomonas, I think we should do it in coexpression space (not correlation over these models). Cooperonic genes will probably create different patterns in both though. I do wonder how far we want to go down this route though - I think we've done a nice job of contextualizing generic genes within previous ML models, which is sort of what we set out to do.

network_analysis/nbconverted/2_calculate_node_statistics.py
jjc2718: Yeah, Casey and I were talking about this in our meeting yesterday - I agree with your interpretation. We think it could be interesting to see if that bump disappears if we use a coexpression network for these analyses (i.e. if generic genes are only very similar when the similarity is calculated using eADAGE LVs, as opposed to just using gene expression over all genes).

Not sure if we want to go further with this analysis, though, or if what we've done is enough for the current paper - I think it's up to you and Casey. Happy to chat about this over Slack if that would be better.

ajlee21:
I agree, I think for the purposes of the paper we were looking at what we could glean about generic genes and why they were "generic" from the existing models (PLIER and eADAGE). 

I do think it would be an interesting analysis to explore, but probably not for this paper you're right.

PR #72, opened 2021-03-25
Update manuscript figures

ajlee21:
Not much code changes in this PR. This PR is updating figures for the manuscript, specifically:

* Enlarging figure panels that were too small to read
* Changing the coloring (figure 1B, figure 4)
* Simplifying what was shown in figure 5
* Add network figure panels

Please let me know any comments you may have on the formatting or if anything doesn't seem clear


network_analysis/3B_community_detection_louvain.py
```diff
@@ -0,0 +1,202 @@
+# ---
```
jjc2718, line 1: Do you want to generate a Python script for `network_analysis/3A_community_detection_planted_partition.ipynb` too? Or did you not change it because we're only using the Louvain results in the main paper?

jjc2718 (approved):
Looks good to me! Just one small question.

network_analysis/3B_community_detection_louvain.py
ajlee21: Since the results were similar for the two different methods, I just chose to show the louvain one in the paper. Thanks for checking

PR #73, opened 2021-03-29
Add new experiment examining counts before and after simulation

ajlee21:
This PR is following up on https://github.com/greenelab/generic-expression-patterns/pull/71 where we noticed the RNA-seq generic genes are low compared to RNA-seq/array generic genes in RNA-seq data.

Our next hypothesis is that the VAE isn’t adequately accounting for low read counts.
Say we have an experiment with low read depth where a set of genes have 0 expression (i.e. not detectable).
After going through the VAE shifting process, the gene counts might get compressed (i.e. very low gene counts are increased and very high gene counts are decreased) so that the new simulated counts for these originally 0-expressed genes, which should have a high error rate and therefore not found to be DE by DESeq, can be found as DE by DESeq with the artificial shift in values.
If this is the case, we will need a way to re-scale values to account for the read depth differences.

Result here is comparing the actual total read counts for a given sample vs simulated total counts. We also compared the counts per gene. 
image

**Observations:**
* For about half of the cases, here is a horizontal trend that indicates that the variance in the actual total counts is lower compared to the simulated total counts. In other words the sequencing coverage of the actual experiment is consistent while the sequencing coverage of the simulated samples is variable.
* Overall, there are cases where most/all samples have a lower total read count in the simulated experiments compared to the actual experiment (when samples are all on one side of the diagonal). There are also cases where some samples have increased counts in the simulated experiment and some have decreased counts in the simulated experiment (i.e. when sample cross the diagonal). 

**Takeaway:**
* These observations are consistent with our hypothesis that the total counts in the simulated experiment are different compared to the actual experiment and creating the opportunity for genes that to be DE that should not be.

* For example, say our actual experiment has some low sequencing coverage (i.e. total read counts) where a set of genes are not detectable (i.e. these genes have 0 read counts). After going through the VAE shifting process, some samples seem to have increased or decreased sequencing coverage compared to the actual sample. This will result in the same gene artifically having a change in read count due to the differences in sequencing coverage between samples.

* To correct for this we can try to re-scale the decoded counts per sample so that the sum of the simulated counts is equal to the sum of the actual total counts.


explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
```diff
@@ -0,0 +1,227 @@
...
+            # simulated_gene_counts = simulated.iloc[1]
+
+            gene_counts = template_gene_counts.to_frame("Original counts").merge(
+                simulated_gene_counts.to_frame("Simulated counts"),
+                left_index=True,
+                right_index=True,
+            )
+
+            # Plot per sample
+            g = sns.scatterplot(
```
jjc2718, line 148: So each point on this plot is a gene, and each subplot is a sample?

How did you choose the genes to look at here? Are these genes that are generic in RNA-seq but not array, and lowly expressed in the RNA-seq data?

explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
```diff
@@ -0,0 +1,227 @@
...
+                data=gene_counts,
+                x="Simulated counts",
+                y="Original counts",
+                alpha=0.3,
+                ax=axes[i],
+            )
+            # axes[i].hexbin(x=gene_counts["Simulated counts"],
```
jjc2718, line 155: Do you still need this code?

explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
```diff
@@ -0,0 +1,227 @@
...
+    ncols=5,
+    nrows=5,
+    fig_width=20,
+    fig_height=20,
+    num_simulated=25,
+    by_sample_or_gene="sample",
+)
+
+# **Observations:**
+# * For about half of the cases, here is a horizontal trend that indicates that the variance in the actual total counts is lower compared to the simulated total counts. In other words the sequencing coverage of the actual experiment is consistent while the sequencing coverage of the simulated samples is variable.
```
jjc2718, line 204: So for the per-sample plots, what you would normally expect to see is a roughly linear relationship (i.e. the simulated counts shouldn't be exactly the same as the actual counts, but they should roughly have the same trend; higher actual counts should generally lead to higher simulated counts)?

If so, I agree, that doesn't seem to be the case in most of these experiments.

I wonder what these plots look like for randomly sampled genes (i.e. genes that aren't lowly expressed in the RNA-seq data, or even genes that aren't generic) - have you thought at all about looking at that?

explore_RNAseq_only_generic_genes/README.md
```diff
@@ -71,3 +71,23 @@ Expression dataset from Crow et. al. downloaded from https://github.com/Pavlidis
 ![hypothesis_2](https://github.com/greenelab/generic-expression-patterns/blob/master/explore_RNAseq_only_generic_genes/array_expression_dist_gene_groups_highlight.svg)
 
 Looks like the RNA-seq generic genes (light blue) are lowly expressed, but not more than other RNA-seq and array generic genes (dark blue) captured on the array.
+
+## VAE compression
+
+Based on experiment for hypothesis 2, we noticed that the RNA-seq generic genes are low compared to RNA-seq/array generic genes in RNA-seq data.
```
jjc2718, line 77: Small clarification:
```suggestion
Based on experiment for hypothesis 2, we noticed that the RNA-seq generic genes tend to have lower read counts compared to RNA-seq/array generic genes in RNA-seq data.
```

jjc2718 (approved):
Looks good - nice work digging into this!

explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
ajlee21: This was me trying to play around to get density information since many points lay on top of each other, but had some trouble with the binning. So its something I'll leave here for now but add a comment to explain.

explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
ajlee21: So these are total counts summed across all genes, to get a sense for the sequencing coverage per sample that will affect the DESeq, which will give more variance (probability of error) to those genes.

https://genomebiology.biomedcentral.com/articles/10.1186/gb-2010-11-10-r106

ajlee21: I'll update the text in the notebook so this is clear. Its a good question

explore_RNAseq_only_generic_genes/1c_original_vs_simulated_counts.py
ajlee21: So I would expect that the sequencing coverage of samples within a simulated experiment to be very tight (so not much spread horizontally. I would also expected that the sequencing coverage of the simulated experiment to be similar to the actual experiment, so the samples should cluster along the diagonal.

ajlee21: I'm still thinking through what it would mean to plot different subsets. So you can look at total counts across the RNA-seq/array generic genes between actual vs simulated. You can also look at total counts across the RNA-seq only generic genes between actual vs simulated. But I'm not sure how to interpret the results here. 

So in general, it looks like the RNA-seq/array generic genes have higher counts in the actual experiment while RNA-seq only generic genes have lower counts in the actual experiment. You still see difference in the total counts between actual vs simulation for both sets of genes.

ajlee21: In terms of correction, I think that plan right now is to re-scale the simulated experiment counts to be match the actual counts. So the relative gene counts within the simulated vs actual experiment can reflect different experiments. But the total counts will be the same between actual and simulated.

PR #75, opened 2021-04-09
More experiments looking into missing generic genes

ajlee21:
This PR adds a new visualization of the data to examine the reason for genes only being found to be generic by SOPHIE.

This PR is following up on the previous observation that genes that are frequently changed in RNA-seq but not array have distinctly lower gene expression profiles compared to other genes.

The plot below shows the average expression of _RNA-seq only_ generic genes in template vs simulated (panel = simulated experiment). The average simulated gene expression seems to be similar to the template or slightly increased
image

The plot below shows the average expression of _RNA-seq/array generic_ genes in template vs simulated (panel = simulated experiment). The average simulated gene expression seems to be mostly similar to the template, some cases where simulated expression is lower compared to the template
image

**Takeaways:**
* Overall, it appears that lower expression values in the template (which correspond to RNA-seq only generic genes) are changed more than genes with higher values
    * I believe this is likely due to the VAE compression -- possibly the ReLU activation function and/or gaussian constraint in the loss function.
    * Why does this compression not affect genes with higher expression values (RNA-seq/array generic genes) as much? This is probably because these values are closer to the mean expression of the compendium. The compression is probably affecting genes on the outliers of the distribution more.

image

* Why is this compression not seen in the array data?
    * In the array data, the gene expression for the RNA-seq only genes vs RNA-seq/array genes were similar in the array training compendium.
    * Overall the variance in array expression is lower compared to RNA-seq so there isnt' as much compression needed

* So genes with low gene expression in the real experiment are getting a boost/increase after going through VAE (simulated experiment) which allows them to be detected as DE.

**Possible solutions:**
* We don't want lowly expressed genes to get artificially detected as frequently DE.
* Would requiring varying parameters for the activation function and weighting for KL term in the loss function. This will need to be addressed in the future, probably not by this manuscript.


jjc2718 (commented):
Looks good to me!

Nice work investigating this, all your conclusions make sense. I remember in your last PR you talked about rescaling the simulated data from the RNA-seq experiments to be closer to the actual counts, to correct for the change due to the VAE compression - did you decide against doing that?

jjc2718 (approved):

ajlee21:
Thanks for asking. After talking with Casey I think that there was some confusion about what the results mean when we were examining differences in sample counts. I think the plan for now is that we at least have an idea for what is causing the issue and we will need to figure out possible solutions to resolve for the future. I think this will more likely be to change the VAE model, but we may look into some sort of scaling of the data. This will be something that future me needs to think about. I think I'll add this as an issue

PR #77, opened 2021-04-16
Update manuscript figures

ajlee21:
Not much code changes in this PR. This is updating the manuscript figures based on the feedback from co-authors. Changes include:

1. Change "generic" to "common"
2. Update panel 5A provided by Dallas to include crc mutant curve
3. Add supplement to explain experiments looking into RNA-seq only common genes
4. Calculate over-representation of (manually derived background set) Crow common genes in our SOPHIE common genes. This will be used to describe the concordance between SOPHIE derived common genes vs manually derived common genes instead of use R^2 term which doesn't reflect the concordance we observe.


human_cancer_analysis/2_identify_generic_genes_pathways.py
```diff
@@ -496,6 +497,35 @@ def shift_template_experiment_with_metadatafile(
...
+# Hypergeometric test:
+# Given N number of genes with K common genes in Crow et al.
+# SOPHIE identifies n genes as being common
+# What is the probability that k of the genes identified by SOPHIE
+# are also common in Crow et al.? What is the probability of drawing
+# k or more concordant genes?
+num_Crow_genes = shared_ranking.shape[0]
+num_generic_Crow_genes = shared_ranking.query(f"{ref_rank_col}>=80.0").shape[0]
+num_generic_SOPHIE_genes = shared_ranking[
+    shared_ranking["Percentile (simulated)"] >= 80.0
```
jjc2718, line 509: You might consider defining this 80% threshold as a constant at the top of this file, so it's easy to change if you ever wanted to.

human_general_analysis/2_identify_generic_genes_pathways.py
```diff
@@ -296,8 +297,39 @@
...
+# Given N number of genes with K common genes in Crow et al.
+# SOPHIE identifies n genes as being common
+# What is the probability that k of the genes identified by SOPHIE
+# are also common in Crow et al.? What is the probability of drawing
+# k or more concordant genes?
+
+num_Crow_genes = shared_ranking.shape[0]
+num_generic_Crow_genes = shared_ranking.query(f"{ref_rank_col}>=80.0").shape[0]
+num_generic_SOPHIE_genes = shared_ranking[
+    shared_ranking["Percentile (simulated)"] >= 80.0
```
jjc2718, line 312: Same comment about putting the threshold in a constant

pseudomonas_analysis/2_identify_generic_genes_pathways.py
```diff
@@ -291,6 +292,37 @@
...
+shared_ranking.head()
+
+# Hypergeometric test:
+# Given N number of genes with K common genes in GAPE.
+# SOPHIE identifies n genes as being common
+# What is the probability that k of the genes identified by SOPHIE
+# are also common in GAPE? What is the probability of drawing
+# k or more concordant genes?
+num_GAPE_genes = shared_ranking.shape[0]
+num_generic_GAPE_genes = shared_ranking[shared_ranking[ref_rank_col] >= 60.0].shape[0]
```
jjc2718, line 304: Why is this threshold different (60% instead of 80%)?

jjc2718 (approved):
Looks good! Just a few small comments.

pseudomonas_analysis/2_identify_generic_genes_pathways.py
ajlee21: The 60% is based on a manual visual inspection of the correlation plot

ajlee21: This question revealed a bug! Turns out that I didn't provide the proper input for this comparison so the distribution wasn't uniform as it should be. Now that its fixed, the threshold should be 80 in both cases:)

jjc2718: Good to hear!

Issue #69, opened 2021-03-02
Pytest failing


ajlee21:
Pytest locally runs but pytetst on github is failing error says it cannot find raw template file

Maybe something is happening with the paths on github
```
error message: <ipython-input-6-53dedeaf9dca> in <module>
----> 1 assert os.path.exists(raw_template_filename)
```

ajlee21:
Addressed in https://github.com/greenelab/generic-expression-patterns/pull/79

PR #79, opened 2021-04-26
Debug pytest failing

ajlee21:
Previously pytest started failing on github despite pytest running successfully on local machine. 

The error reported by github is described here: https://github.com/greenelab/generic-expression-patterns/issues/69

Pytests were running successfully on github and we didn't make any changes to the component of the scripts that the error is referencing.

Previously I was not able to determine the reason for this error so I temporarily commented out the pytest calls that were failing. Here I am going back to determine the cause and try to fix it

Some possibilities I looked into were:
1. If the directories existed that are being written to, which looks to be the case
2. If this was an issue with Rmagic, but if that were the case then cells in `2_identify_generic_genes_pathways.ipynb` would also fail which wasn't the case

My solution: Add the missing file and remove the section to test downloading the data. I am still not exactly clear on the reason for this failure, but my suspicion is that perhaps there is an issue with accessing recount to download data not playing well with github actions


ben-heil (commented):
Do you ever make the data directory to store the recount2 results files in? It seems like the test may be downloading the data but not having a place to store it. The only other reference to the data dir is at the bottom of `1_process_recount2_data`, but it looks like you decided not to use the file there

ajlee21:
Here is the line for where I made the directory that the data should be writing into: https://github.com/greenelab/generic-expression-patterns/blob/573c99cbcea8233e8ef7ee1f2296b01d83c12bf7/tests/1_process_recount2_data.py#L75

That was my initial thought too so I tested to see if the `project_dir` exists and it says it did exist within the function `get_recount2_template_experiment` but then the assertion for `raw_template_filename` would fail in the notebook

ben-heil (approved):
Huh, I don't know what's causing it then. Went ahead and approved the PR

ajlee21:
Thank you for thinking about it..i am also at a loss..thus all the failed commits above:) This code also was working previously and stopped working a couple months ago but I didn't update any of these functions in that time so its very curious

PR #80, opened 2021-05-18
Update figures with new acronym

ajlee21:
This PR is updating figures (figure 4, S1) to use new acronym "common DEGs" to refer to commonly differentially expressed genes. These genes were previously referred to as "generic genes".

There is no large code changes to be reviewed here.


jjc2718 (approved):
Looks good to me!

PR #81, opened 2021-05-20
More figure updates

ajlee21:
This PR is adding some additional  edits to figures based on co-author suggestions. These edits include:

1. Formatting text from log10 --> log_10
2. Add the dataset that SOPHIE trained on to the axis labels to help be more clear
3. Text was very small for Figure 3D, so created a new heatmap plot and pulled this panel into the supplement


other_enrichment_methods/2_viz_pathway_ranking.py
```diff
@@ -0,0 +1,596 @@
...
+    fig += pn.geom_point()
+    fig += pn.geom_point(
+        plot_df[plot_df["Percentile rank"] > 90.0],
+        pn.aes(x="Test statistic", y="Percentile rank"),
+        color="red",
+    )
+
+    # Only used for array data
+    # Manually stagger labels since they overlap here
+    """
```
jjc2718, line 157: Are you using this function for array data at all? If so, you might want to pass in an argument for array or RNA-seq, then run the relevant code block.

jjc2718 (approved):
Looks good! Two small questions.

other_enrichment_methods/2_viz_pathway_ranking.py
ajlee21: This section is commented out since we only decided to show the results for the RNA-seq data since the array didn't produce a curve for ORA because there were some simulated experiments that didn't find any DEGs. I'll go ahead and add this params for future use

PR #85, opened 2021-07-06
Compare common vs other gene expression

ajlee21:
This PR performs the experiment discussed at Greenelab retreat: Are these common DEGs common because of some technical source? 

To answer this question, we compared the average expression of common DEGs vs other genes. Based on our analysis, there doesn't appear to be a difference in expression between common genes vs other genes for this analysis. It would be interesting to continue to followup on this analysis to try to rule out other possible factors that could explain why these common DEGs are so common (for future work).


nrosed:
Interesting! I'm just so surprised by this because when you do DE analyses, you find that the results seem to be well-calibrated (in non-cancer settings). However, this suggests significant inflation that isn't observed in a single experiment. Therefore, there must exist model misspecification in the classical DE set-up, which isn't too surprising since genes are interrelated and this correlation is not taken into account. This is a really nice way of proving this model misspecification!

Unrelated questions: do you find that your results are well-calibrated, i.e. the genes you find to be overall DE after simulation follow the qq-plot?

If you are interested, I think another direction to see if there are additional technical artifacts would be to look into an MA plot (Bland–Altman plot) to see if it has to do with some strange relationship between the mean and variance. 

Another plot would be to look at the relationship between inter-replicate and intra-replicate variance. I'm assuming that if they are highly reactive, then they should have a high coefficient of variation across all experiments, but maintain a low coefficient of variation within any particular set of replicates.

ajlee21:
Thank you so much for your insight/questions and suggestions. I'll open up issues for these

> Interesting! I'm just so surprised by this because when you do DE analyses, you find that the results seem to be well-calibrated (in non-cancer settings). However, this suggests significant inflation that isn't observed in a single experiment. Therefore, there must exist model misspecification in the classical DE set-up, which isn't too surprising since genes are interrelated and this correlation is not taken into account. This is a really nice way of proving this model misspecification!
> 
> Unrelated questions: do you find that your results are well-calibrated, i.e. the genes you find to _not_ be overall DE after simulation follow the qq-plot?

Just to clarify, are you suggesting to look at the qqplot (real vs simulated) of a _not_ common DEGs to compare the distribution of this non common DEG in the real vs simulated experiment? So this will allow us to look at how dispersed/skewed the expression is for a non common DEGs before and after simulation? And the hope is that we will have more information about how the simulation/VAE is affecting these non common DEGs. So its similar to what I already did but here it would be looking at individual genes instead of all the non common DEGs together?

> If you are interested, I think another direction to see if there are additional technical artifacts would be to look into an MA plot (Bland–Altman plot) to see if it has to do with some strange relationship between the mean and variance.

Thanks I'll have to look at that for common DEGs vs other non common ones

> Another plot would be to look at the relationship between inter-replicate and intra-replicate variance. I'm assuming that if they are highly reactive, then they should have a high coefficient of variation across all experiments, but maintain a low coefficient of variation within any particular set of replicates.

I think that would be an interesting experiment. I might have misunderstood something here. So we would expect a common DEGs to have high CV within the experiment (i.e. high variance between samples within the same experiment since the gene is found to be DE) but low CV for samples across experiments (i.e. low variance between samples from different experiments because the gene is equally DE in two different experiments).

nrosed:
> Just to clarify, are you suggesting to look at the qqplot (real vs simulated) of a _not_ common DEGs to compare the distribution of this non common DEG in the real vs simulated experiment? So this will allow us to look at how dispersed/skewed the expression is for a non common DEGs before and after simulation? And the hope is that we will have more information about how the simulation/VAE is affecting these non common DEGs. So its similar to what I already did but here it would be looking at individual genes instead of all the non common DEGs together?

What I am curious about is looking at the genes that were not found to be differentially expressed by your method. Essentially, I am curious if SOPHIE is better calibrated than a regular differential expression method (DESeq). So I am curious about comparing the qq-plots of p-values from SOPHIE and DESeq of genes not found to be differentially expressed.  Under the null, they should follow the uniform distribution, but I bet that SOPHIE is better calibrated (has less inflated p-values) than DESeq. This isn't really related to understanding common DEGs, but just comparing the results for a individual experiment between SOPHIE and DESeq. Does this make sense? Maybe you have done something similar but I missed it? This is totally just a curiosity from me, so please ignore if not interested.

> > Another plot would be to look at the relationship between inter-replicate and intra-replicate variance. I'm assuming that if they are highly reactive, then they should have a high coefficient of variation across all experiments, but maintain a low coefficient of variation within any particular set of replicates.
> 
> I think that would be an interesting experiment. I might have misunderstood something here. So we would expect a common DEGs to have high CV within the experiment (i.e. high variance between samples within the same experiment since the gene is found to be DE) but low CV for samples across experiments (i.e. low variance between samples from different experiments because the gene is equally DE in two different experiments).

I drew out my thinking below, but your logic also works. My thinking was that commonDEGs are all over the place across experiments and generally have a high CV, but to be truly differential, they somehow have a lower within-condition CV. 
photo_2021-07-09 11 23 33

Issue #84, opened 2021-06-29
Comments provided by Tessa (first SOPHIE user)


ajlee21:
Below are comments provided by Tessa:

- Sklearn version. 0.23.2

- I commented out gxx_linux-64 from yml file because it was stopping my install. I’m pretty sure that’s something you would only need to install in order to use it to install other stuff with pip, and I didn’t run into any other issues starting up from the yml, so I left it as-is, but it could be that’s contributing to some of my issues. 

- Made it through step 8! For step 9, i got an error FileNotFoundError:

 [Errno 2] No such file or directory: '/Users/tgreen/Dropbox (HMS)/Tessa_DSM/AERD/Projects/Broad-aspirin/sophie/pseudo_experiment/template_normalized_data_aspirin-all_test.txt'

I created a pseudo_experiment directory and ran it again, which got me further. 

- Additional issues — the notebook doesn’t say that the groups .tsv needs to be in the folder /data/metadata, I had them in the /data folder but it’s all good. 

- I installed the package and then have been using a different folder for my analysis. I had to copy over the generic_expression_pattern_modules folder to get through the R bit. Have hit on a new issue: 

 cannot open file '/Users/tgreen/Dropbox (HMS)/Tessa_DSM/AERD/Projects/Broad-aspirin/sophieDE_stats/DE_stats_template_data_aspirin-all_real.txt': No such file or directory

Error in file(file, ifelse(append, "a", "w")) : 
  cannot open the connection

I suspect this is something going wonky with R not automatically adding slashes into its equivalent of os.path.join in the back end (I’m more comfortable with R than python, and I was very confused last week when I discovered that os.path.join did this!). Mostly thinking that because there’s no slash between Sophie and DE_stats. I went in and modified the DE_analysis.R file to add those slashes in, and now it works. 

- In the second of the R boxes I had to modify line 9 to read "        "/pseudo_experiment/selected_simulated_data_",
because it was trying to save in a directory sophiepseudo_experiment (which doesn’t exist).

tessadgreen:
For the slashes--I will need to test it, but I think the issue was that I didn't have a "/' at the end of my "local_dir" in the config file. Python didn't care but R needs it. I fixed the issue by going in and modifying your code, but fixing the config is definitely easier/better

PR #87, opened 2021-07-09
Address user issues

ajlee21:
This PR is meant to address issues that arose when a user tried to run SOPHIE. The issues are described here: https://github.com/greenelab/generic-expression-patterns/issues/84

The main issues include:
1. Adding a version for sklearn in order to use pre-trained models that we pickled
2. Adding clarification to the documentation about the expected input format
3. Creating two different environments, one for linux and another for mac


environment_linux.yml
```diff
@@ -16,7 +16,7 @@ dependencies:
 - bioconda::bioconductor-DEFormats
 - bioconda::bioconductor-clusterprofiler
 - conda-forge::python=3.7
-- conda-forge::scikit-learn
+- conda-forge::scikit-learn=0.23.2
```
ben-heil, line 19: You may want to explicitly set the pickle version as well. It _shouldn't_ change since you have explicit versions for sk-learn and pandas, but explicitly setting it would make sure that it is correct

environment_mac.yml
```diff
@@ -0,0 +1,44 @@
+name: generic_expression_mac
```
ben-heil, line 1: Is the difference between the mac and linux environment files just that the mac env leaves out the gxx dependency? If that's the case it may be easier to use a single environment and just not list gxx since your compiler will change between operating systems and (maybe) processors anyway. 

If there's a reason that you are setting gxx explicitly though then the solution you're using is probably the best one

new_experiment/find_specific_genes_in_new_experiment.py
```diff
@@ -55,6 +55,8 @@
 # The remaining parameters specify filenames that intermediate data files will be written to.
 #
 # The user also needs to provide metadata files:
+#
+# These files should be located in `data/metadata/` directory.
```
ben-heil, line 59: You may also want to add an empty file to `data/metadata` called `PLACE_GROUPS_FILE_HERE` or something

ben-heil (approved):
Looks good! I left a couple quality of life comments that you can take or leave. You also may want  to add the trailing slashes (or a comment about them) to the config file to avoid [this issue](https://github.com/greenelab/generic-expression-patterns/issues/84#issuecomment-870980186) in the future

environment_linux.yml
ajlee21: Good point, i'll add this as well.

environment_mac.yml
ajlee21: This was another legacy package that I have, so I'm not sure if its required. I'll try to test this -- yay for github actions.

new_experiment/find_specific_genes_in_new_experiment.py
ajlee21: Good idea!

PR #90, opened 2021-08-02
Crow et al model

ajlee21:
This PR is training VAE on the Crow et al dataset (human array mixed bag expression dataset) in order to check that array-trained SOPHIE model can recapitulate manual results using an array dataset. This PR is in response to reviews from the SOPHIE manuscript where we compared SOPHIE trained on RNA-seq (recount2) vs manually curated array dataset without first comparing SOPHIE trained on array data vs manually curated array dataset.

Most of the code is the same as previously reviewed but we created a new instance of it, so no need for a very heavy code review here. 

The results look consistent with what we've seen previously, but there are a few things I'll want to check when I return from brring OOO:
1. The training curve looks a odd to me - i.e. pretty flat, making me think its not learning. So I'll want to try to play with the params a bit more
2. There were some negative values in the raw data downloaded and I want to look into this and make sure I have handled the data correctly.


expression_common_vs_other/1_expression_common_vs_other.py
```diff
@@ -0,0 +1,164 @@
...
+        base_dir, dataset_name, params["reference_gene_filename"]
+    )
+else:
+    DE_prior_filename = params["reference_gene_filename"]
+ref_gene_col = params["reference_gene_name_col"]
+ref_rank_col = params["reference_rank_col"]
+
+figure_filename = f"gene_ranking_{col_to_rank_genes}_tmp.svg"
+
+corr, shared_ranking = ranking.compare_gene_ranking(
```
jjc2718, line 88: This is exclusively array data you're plotting here, right? Interesting that you still see the group of genes in the bottom right, that went away in your last PR when you applied MR normalization to the RNA-seq data.

What kind of normalization are you doing for the array data? Do you think it could be affecting the cluster of genes in the bottom right here as well? Or is this notebook using RNA-seq data?

human_general_array_analysis/1_process_data.py
```diff
@@ -0,0 +1,242 @@
...
+)
+
+# Check normalized data values
+normalized_compendium = pd.read_csv(
+    normalized_compendium_filename, sep="\t", index_col=0
+)
+
+normalized_compendium.head(10)
+
+sns.displot(normalized_compendium["AA06"])
```
jjc2718, line 180: Is this distribution fairly typical of what you're seeing in other genes, or is there a reason you picked AA06 here?

human_general_array_analysis/2_identify_generic_genes_pathways.py
```diff
@@ -0,0 +1,732 @@
...
+
+# +
+# Get generic genes identified by Crow et. al.
+DE_prior_filename = params["reference_gene_filename"]
+ref_gene_col = params["reference_gene_name_col"]
+ref_rank_col = params["reference_rank_col"]
+
+figure_filename = f"gene_ranking_{col_to_rank_genes}.svg"
+
+corr, shared_ranking = ranking.compare_gene_ranking(
```
jjc2718, line 513: This plot actually looks more well-behaved to me than most of the RNA-seq plots you've made in that most values are fairly close to the diagonal. This one is using array data to train SOPHIE, right?

jjc2718 (approved):
Looks good to me!

When you say this:

> The training curve looks a odd to me - i.e. pretty flat, making me think its not learning. So I'll want to try to play with the params a bit more

Do you mean [this file](https://github.com/greenelab/generic-expression-patterns/blob/e8e42cdaea644401b1e26d1512e48b89fb69663f/human_general_array_analysis/logs/NN_2500_30/tybalt_2layer_30latent_hist.svg)? It could just be the scale that's making it look flat; i.e. the first epoch has a really high loss which makes the y-axis scale too large to see differences in subsequent epochs.

Do you know of examples of other learning curves (in this repo or elsewhere) that look better to you?

expression_common_vs_other/1_expression_common_vs_other.py
ajlee21: This is comparing common findings from manually-curated array experiment results vs SOPHIE trained on RNA-seq data.

human_general_array_analysis/2_identify_generic_genes_pathways.py
ajlee21: Yes, here we trained SOPHIE on array data from Crow et al and then compared the common findings to those found by manually curating experiments from Crow et al.

human_general_array_analysis/1_process_data.py
ajlee21: This was more me testing what the distribution of the data looked like after normalization - to make sure the values were between 0 and 1

ajlee21:
> Looks good to me!
> 
> When you say this:
> 
> > The training curve looks a odd to me - i.e. pretty flat, making me think its not learning. So I'll want to try to play with the params a bit more
> 
> Do you mean [this file](https://github.com/greenelab/generic-expression-patterns/blob/e8e42cdaea644401b1e26d1512e48b89fb69663f/human_general_array_analysis/logs/NN_2500_30/tybalt_2layer_30latent_hist.svg)? It could just be the scale that's making it look flat; i.e. the first epoch has a really high loss which makes the y-axis scale too large to see differences in subsequent epochs.
> 
> Do you know of examples of other learning curves (in this repo or elsewhere) that look better to you?

Here is the loss curve for SOPHIE trained on RNA-seq data (https://github.com/greenelab/generic-expression-patterns/blob/master/human_general_analysis/logs/NN_2500_30/tybalt_2layer_30latent_hist.svg)

Here is the loss curve for SOPHIE trained on a different array dataset (Powers et al): https://github.com/greenelab/generic-expression-patterns/blob/master/human_cancer_analysis/logs/NN_2500_30/tybalt_2layer_30latent_hist.svg

I understand that the reason the curve appears more flat is due to the difference in axis range, but I guess I wasn't expecting there to be a large difference between the training vs validation loss values (similar to what you see in the loss curves for SOPHIE trained on RNA-seq data). Though it looks like the loss curves for SOPHIE trained on Powers et al array data is similar to SOPHIE trained on Crow et al array data. So perhaps this loss curve is as expected.

PR #94, opened 2021-09-16
Update workflow figure

ajlee21:
The main change in this PR is updating the workflow to describe SOPHIE so that its clear to the reader Figure 1A. Other workflow panels were removed from figures to avoid confusion with have multiple workflows and instead the text will describe what we did. If you have any thoughts on this figure, please let me know.

Minor updates to the readme to align with the edited manuscript as well as text updates to the comments in notebooks.


jjc2718 (approved):
Looks good to me!

I like the new workflow figure! One small idea: if think you have room maybe you could show the multiple perturbations that generate the multiple simulated experiments? Right now you just have one shift of the template experiment in the latent space, but you have a bunch of simulated experiments - it might not be clear to unfamiliar readers that each shift corresponds to a different simulated experiment.

I guess this is described in the ponyo paper, though, so if you're pointing to that paper already maybe you don't need this much detail.

ajlee21:
> Looks good to me!
> 
> I like the new workflow figure! One small idea: if think you have room maybe you could show the multiple perturbations that generate the multiple simulated experiments? Right now you just have one shift of the template experiment in the latent space, but you have a bunch of simulated experiments - it might not be clear to unfamiliar readers that each shift corresponds to a different simulated experiment.
> 
> I guess this is described in the ponyo paper, though, so if you're pointing to that paper already maybe you don't need this much detail.

We are trying to off load the description of the simulation and how it works to the ponyo paper to make this more readable. However I do see your point. I can try to play with this a bit more. Thank you for the feedback

Issue #97, opened 2021-09-24
Update analysis notebooks to use newest version of ponyo


ajlee21:
The using the current implementation of ponyo assumes only two fixed types of metadata files are provided for the input compendia. So if a user wants to try a new VAE model, they will need to use the newest version of ponyo to be able to input a new metadata file. I have not tested if updating ponyo would break other components of the SOPHIE analysis pipeline. This is something that should be addressed to continue to make SOPHIE more usable

ajlee21:
Addressed in https://github.com/greenelab/generic-expression-patterns/pull/98

PR #98, opened 2021-09-29
Update to latest version of ponyo

ajlee21:
Currently, the way users can apply SOPHIE to their own data is to use an existing pre-trained VAE model and input their own template experiment to then simulate data.

However, its also possible that users may want to train a new VAE model and use a template experiment from their training dataset to simulate data. This user case isn't currently possible since SOPHIE is using an older version of ponyo, which has a fixed metadata file it expects to to get as input when simulating data  using a template experiment that is contained within the training compendium. Ponyo has since been updated to allow users to input their own metadata file. In this PR I am testing to see if I can use the latest version of ponyo to allow users this option. If this is possible I will need to clean up the old analysis notebooks that perform a workaround where I've manually copy-and-pasted the new version of ponyo code into my notebooks.

This PR addresses https://github.com/greenelab/generic-expression-patterns/issues/97

The main changes can be found in:
1. Readme
2. `new_model_experiment/`

Many other files were updated now with the new version of ponyo and are much cleaner and don't need much review


jjc2718 (approved):
Looks good!

PR #102, opened 2022-01-11
Add a couple of small analysis based on reviewer comments

ajlee21:
This PR adds a couple of evaluation analyses based on reviewer comments: to prepare for re-submission

1. Varying the number of simulated experiments to show that 25 is a sufficient number of experiments to use as a null set (New figure:`figure_generation/output/figure_S4.svg`)
2. Directly calculating the overlap of the top commonly DEGs across platform and context to show that there are some consistent common DEGs (New notebook: `compare_experiments/consistency_common_DEGs.ipynb` and figure: `figure_generation/output/figure_S2.svg`)

Other changes were minor and me cleaning up old figures.


compare_experiments/consistency_common_DEGs.py
```diff
@@ -0,0 +1,212 @@
...
+    "cross_all_venn.svg",
+    format="svg",
+    bbox_inches="tight",
+    transparent=True,
+    pad_inches=0,
+    dpi=300,
+)
+# -
+
+# It looks like the majority of the top 20% commonly changed genes are not consistent across all compendia, but there are some genes that are consistent across platform and context (consistent across 2 compendia but not all 3)
```
jjc2718, line 210: This is interesting - I wonder how much overlap you'd expect by chance. You'd have to think carefully about your null model though, since probably not all genes are equally likely to be DE in general. Depends on how convinced you think reviewers will be by these plots (the common genes seem to be somewhat consistent but not _super_ consistent/obvious).

jjc2718 (approved):
Looks good! A couple of small comments.

compare_experiments/consistency_common_DEGs.py
```diff
@@ -0,0 +1,212 @@
...
+import os
+import seaborn as sns
+import pandas as pd
+import numpy as np
+import matplotlib
+import matplotlib.pyplot as plt
+from matplotlib_venn import venn2, venn3
+# -
+
+threshold = 80
```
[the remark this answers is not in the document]
ajlee21: This cutoff will vary depending on the experiment and something I was playing around with just to see if consistencies changed. So I think I'll leave this variable in the header for now

compare_experiments/consistency_common_DEGs.py
ajlee21: This is a very good question. The reviewer that mentioned this, seemed accept our approach, but was curious about the boundaries of SOPHIE, specifically how robust common DEGs were across compendia. I think it makes sense that there are some "general common DEGs" that are robust across context and compendia. But there are differences that I would guess are due to the different contexts or technical differences in platform. We tried to do a little digging into the technical differences that found some genes to be common in only the RNA-seq setting. 

I definitely think that its a very good question: How robust are common DEGs across different types of compendia? What accounts for the difference in some genes being common in one compendia vs another? I don't think that this is the paper to really explore those questions. So this analysis as well as our RNA-seq only common DEGs analysis that is another supplemental are peaks into starting to look into that. And I agree that in order to do this more explicitly in the future, we would need to think carefully about a control - one idea i had for a possible negative control is to use specific genes, though perhaps that's circular.

jjc2718: Got it - I'm convinced! Hopefully your reviewers feel the same 🙃

PR #103, opened 2022-02-02
Minor updates to figures

ajlee21:
This PR makes a couple of minor updates to figures in preparation for re-submission

1. In figure 5, a new pathway schematic was provided my collaborators to include the downstream effects on arginine metabolism. The knockout genes (cbrB and crc) were removed from the plots in panels B and C since they are "outliers" and the scaling for the z-score could be illustrated more clearly.
2. In figure 3 I rearranged the layout of the panels to minimize dead space and make each panel larger and more readable


pseudomonas_analysis/viz_E-GEOD-33245_cbrB_v_WT.py
```diff
@@ -242,8 +242,19 @@ def make_volcano_template_highlight_genelist(
 cbrB_crc_df.loc[cbrB_crc_df["Gene Name"] == "cbrB", "gene group"] = "cbrB"
 cbrB_crc_df.loc[cbrB_crc_df["Gene Name"] == "crc", "gene group"] = "crc"
 
+print(cbrB_crc_df.shape)
 cbrB_crc_df.head()
 
+# +
+# Drop cbrB and crc rows for plot
```
jjc2718, line 249: Maybe you could comment why you're doing this here? Something like what you said in the PR description would be fine, just so you remember in the future why you dropped these rows.

jjc2718 (approved):
Looks good! One small suggestion.

pseudomonas_analysis/viz_E-GEOD-33245_cbrB_v_WT.py
ajlee21: Very good idea. I'll add this

PR #122, opened 2022-08-01
Reorder figures based on mention in text

ajlee21:
This PR contains the following minor changes:

1. Italicize gene names in figure 1A
2. Re-ordering the supplemental figures based on when they are mentioned in the text. The order is changed now that the paper was re-formated to be a methods paper.


jjc2718 (approved):
LGTM!

PR #123, opened 2022-09-08
Add stats test based on reviewer comment

ajlee21:
Final edit to SOPHIE manuscript before its acceptance - A reviewer asked us to quantify the gain in using SOPHIE that is depicted in Figure 1D. So I have added a Wilcoxon signed rank test to compare how common genes change rank between traditional vs SOPHIE. And similarly using specific genes.


explore_simulation_approach/3_evaluate.py
```diff
@@ -33,6 +33,7 @@
 import pandas as pd
 import seaborn as sns
 import numpy as np
+import scipy
```
jjc2718, line 36: Would this work?
```suggestion
import scipy.stats
```
Or `from scipy.stats import wilcoxon` would be fine too - it's just that`scipy` is a very large library so it probably isn't necessary to import the whole thing here, maybe better to stick with just the module or function you need.

jjc2718 (approved):
Looks good to me! Congrats on the acceptance 🎉

PR #125, opened 2023-01-20
Edit figures for GBP final proof

ajlee21:
This PR addresses final edits to figures based on editor comments


cgreene (approved):
Changes look fine to me - note that there is a failing dependency - looks like some issues in conda.

ajlee21:
Looks like there is some issue with setting up the conda env using github actions, but the conda env installs successfully locally. There doesn't seem to be a way to access/debug the environment that github actions is using. This issue seems to be known with the only solution being to uninstall/reinstall conda, which I don't think I can do on github actions. So for now, I have commented out the github actions tests.

https://github.com/conda/conda/issues/12155

